PR #1, opened 2016-05-31
Thellimist/tweaks

thellimist:


config.js
```diff
@@ -1,5 +1,5 @@
 {
-    "debug": false,
+    "debug": true,
```
jwineman: We want the default config value for `debug

PR #4, opened 2016-06-01
Thellimist/proxy.php

thellimist:


cloudflare2.php
```diff
@@ -14,7 +16,8 @@
 ?>
 <div id="root" class="cloudflare-partners site-wrapper"></div>
 <script>
-var absoluteUrlBase = '<?=plugins_url('/cloudflare/');?>'
+var absoluteUrlBase = '<?=plugins_url('/cloudflare/');?>';
+cfCSRFToken = '<?=CF\SecurityUtil::csrfTokenGenerate($wordpressAPI->getHostAPIKey(), $wordpressAPI->getUserId());?>';
```
jwineman: `getHostAPIKey()` returns null cause this plugin doesn't use the host API.  I don't know what we're going to use for the secret yet but need to log this as a TODO to follow up on.

proxy.php
```diff
@@ -1,12 +1,22 @@
 <?php
 require_once("vendor/autoload.php");
+
+// include wp-load.php relatively
+// $parse_uri = explode( 'wp-content', $_SERVER['SCRIPT_FILENAME'] );
+// require_once( $parse_uri[0] . 'wp-load.php' );
```
jwineman: Lets remove debug comments

proxy.php
```diff
@@ -15,17 +25,34 @@
...
-$isCSRFTokenValid = (($request->getMethod() === "GET") ? true : \CF\SecurityUtil::csrfTokenValidate($cpanelAPI->getHostAPIKey(), $cpanelAPI->getUserId(), $request->getBody()['cfCSRFToken']));
+$isCSRFTokenValid = false;
+$isCSRFTokenValid = ($request->getMethod() === "GET") ? true : CF\SecurityUtil::csrfTokenValidate($wordpressAPI->getHostAPIKey(), $wordpressAPI->getUserId(), $request->getBody()['cfCSRFToken']);
 unset($body['cfCSRFToken']);
 $apiResponse = "";
+$apiRouter = null;
+
+if(isClientAPI($request->getUrl())) {
+	$apiRouter = new CF\Router\DefaultRestAPIRouter($wordpressIntegration, $clientAPIClient, $clientAPIClientRoutes);
+} else {
```
jwineman: Add a comment about the fact that this is routing plugin API requests.

PR #6, opened 2016-06-02
Thellimist/zones

thellimist:


src/Test/ClientActionsTest.php
```diff
@@ -0,0 +1,52 @@
...
+        $this->mockWordPressAPI->method('getDomainList')->willReturn(
+            array($wordPressDomain)
+        );
+        $this->mockClientAPI->method('responseOk')->willReturn(true);
+        $this->mockClientAPI->method('callAPI')->willReturn(array("result" => array()));
+        $response = $clientActions->returnWordPressDomain();
+
+        $this->assertEquals($wordPressDomain, $response["result"][0]["name"]);
+    }
+}
```
thellimist: Since I don't know what "get_site_url" returns when in development mode (localhost), a domain does not exist (pure ip) or wordpress hosted; we may need to add other tests

src/Wordpress/WordPressAPI.php
```diff
@@ -53,7 +53,11 @@ public function getHostAPIKey() { return null; }
      * @param null $userId
      * @return mixed
      */
-    public function getDomainList($userId = null) { return null; }
+    public function getDomainList($userId = null)
```
jwineman: This will only ever return one domain, lets rename it to getDomain() or something that emphasizes the fact that it doesn't return a list.

src/Test/ClientActionsTest.php
jwineman: Yeah so for ClientActionsTest we just need to test that it merges the results of (mocked) getDomainList() with the API response.  In WordPressAPITest test that getDomainList() returns an array of size one with a string.

PR #7, opened 2016-06-07
Thellimist/accounts

thellimist:


src/API/Plugin.php
```diff
@@ -0,0 +1,25 @@
+<?php
+
+namespace CF\API;
+
+class Plugin extends Client
+{
+    const PLUGIN_API_NAME = 'PLUGIN API';
```
thellimist: instead of PLUGIN_API_NAME, CLIENT_API_NAME, HOST_API_NAME we can use API_NAME to represent all of them in AbstractAPIClient.php

src/API/AbstractAPIClient.php
```diff
@@ -30,24 +29,29 @@ public function __construct(IntegrationInterface $integration)
...
      * @param Request $request
+     *
      * @return array|mixed
      */
     public function callAPI(Request $request)
     {
         try {
+            if ($this->getAPIClientName() == 'PLUGIN API') {
+                $this->logger->error('Plugin API can not be called');
+            }
```
jwineman: This is implementation specific logic which we want to keep out of the abstract class.  Instead of implementing this logic here lets extend callAPI in CF\API\Plugin:
```
public function callAPI(Request $request) {
   return $this->createAPIError("The url: " . $request->getUrl() . " is not a valid path.");
}
```

src/Wordpress/PluginActions.php
```diff
@@ -0,0 +1,55 @@
...
+
+        // TODO: what should the exact response be?
+        $response = array(
+            'success' => 'true',
+            'result' => array(
+                    'email' => $email,
+                ),
+            );
+
+        $this->dataStore->createUserDataStore($apiKey, $email, null, null);
```
jwineman: [update_option() ](https://codex.wordpress.org/Function_Reference/update_option) returns a boolean based on whether the call succeeded or not.  Modify `CF\Wordpress\Datastore::createUserDataStore()` to return true/false based on whether both `update_option()`s succeed and then wrap this call in an if statement that only returns the response if it was successful.  If it was not successful return a `$api->createAPIError()` explaining that you were unable to save the user's credentials.

src/Wordpress/WordPressAPI.php
```diff
@@ -75,8 +75,7 @@ public function getHostAPIKey()
      */
     public function getDomainList($userId = null)
     {
-        // TODO: fix this
-        return array('example.com');//get_site_url();
```
jwineman: Do we need this comment?

src/Wordpress/WordPressAPI.php
```diff
@@ -75,8 +75,7 @@ public function getHostAPIKey()
      */
     public function getDomainList($userId = null)
     {
-        // TODO: fix this
```
jwineman: Is this fixed?

src/Wordpress/PluginActions.php
```diff
@@ -0,0 +1,55 @@
...
+        $this->logger = $defaultIntegration->getLogger();
+        $this->request = $request;
+    }
+
+    /**
+     * GET /zones.
+     *
+     * @return mixed
+     */
+    public function loginWordPress()
```
jwineman: Write tests for this function please.

proxy.php
```diff
@@ -32,30 +31,32 @@
...
+    $apiRouter = new CF\Router\DefaultRestAPIRouter($wordpressIntegration, $pluginAPIClient, $pluginAPIPluginRoutes);
 }
 
-if($isCSRFTokenValid) {
+if ($isCSRFTokenValid) {
     $apiResponse = $apiRouter->route($request);
 } else {
-    $apiResponse = $apiRouter->getAPIClient()->createAPIError("CSRF Token not valid.");
+    $apiResponse = $apiRouter->getAPIClient()->createAPIError('CSRF Token not valid.');
     // TODO: We should add return error response here.
```
jwineman: Is this still a TODO?

src/Wordpress/PluginActions.php
```diff
@@ -0,0 +1,55 @@
...
+    {
+        $apiKey = $this->request->getBody()['apiKey'];
+        $email = $this->request->getBody()['email'];
+
+        // TODO: what should the exact response be?
+        $response = array(
```
jwineman: This is the correct response, remove the comment but move this response structure into CF\API\Plugin::createAPISuccessResponse($result) 
```
$response = array(
            'success' => 'true',
            'result' => $result
            );
```

src/Test/PluginActionsTest.php
```diff
@@ -0,0 +1,93 @@
...
+
+        $request = new Request(null, null, null, null);
+        $request->setBody(
+            array(
+                'apiKey' => $apiKey,
+                'email' => $email,
+            )
+        );
+
+        $this->mockPluginAPI->method('createAPISuccessResponse')->willReturn(
```
jwineman: You don't need to mock this because it will never be called right?

src/Test/PluginTest.php
```diff
@@ -0,0 +1,52 @@
+<?php
```
jwineman: Test namespace should match the regular namespace plus "/Test" so if you have
`CF\API\Plugin` the namespace of its test should be `CF\Test\API\Plugin`

src/Test/PluginActionsTest.php
thellimist: in PluginActions.php
`$response = $this->api->createAPISuccessResponse(array('email' => $email));`
is being called so the response is directly relevant with this method

src/API/Plugin.php
```diff
@@ -0,0 +1,45 @@
...
+    {
+        return self::PLUGIN_API_NAME;
+    }
+
+    /**
+     * @param Request $request
+     *
+     * @return array|mixed
+     */
+    public function callAPI(Request $request)
```
jwineman: Write a test that ensures this always returns an error please.

src/Test/PluginActionsTest.php
thellimist: Oww, if you mean `request` you're right

jwineman: This test case is testing the failure scenario right?  If its supposed to fail then you'll never call `createAPISuccessStory()` as part of this test so I don't thin you need to mock it.  I could be wrong.

src/Test/API/PluginTest.php
```diff
@@ -49,4 +50,18 @@ public function testCreateAPISuccessResponse()
...
+    public function testCallAPI()
+    {
+        $request = new Request(null, null, null, null);
+        $pluginAPI = new Plugin($this->mockDefaultIntegration);
+
+        $this->mockWordPressAPI->method('createAPIError')->willReturn($error);
+
+        $response = $pluginAPI->callAPI($request);
+
+        var_dump($response);
```
jwineman: Get rid of the var_dump()

PR #11, opened 2016-06-17
Thellimist/merged

thellimist:


src/Wordpress/DataStore.php
```diff
@@ -9,6 +9,7 @@ class DataStore implements DataStoreInterface
 {
     const API_KEY = 'cloudflare_api_key';
     const EMAIL = 'cloudflare_api_email';
+    const IP_REWRITE = 'cloudflare_ip_rewrite';
```
jwineman: We don't want to add a new const every time we add a setting so lets do this instead.
```
const CLOUDFLARE_SETTING_PREFIX = "cloudflare_";
const IP_REWRITE = "ip_rewrite";
```

src/Wordpress/DataStore.php
```diff
@@ -70,4 +71,26 @@ public function getCloudFlareEmail()
...
     }
+
+    /**
+     * @return (bool)
+     */
+    public function getIpRewrite()
+    {
+        return get_option(self::IP_REWRITE);
+    }
+
```
jwineman: Then this can become 
```
public getPluginSetting($settingId) {
    return get_option(self::CLOUDFLARE_SETTING_PREFIX + $settingId);
}
```

src/Wordpress/DataStore.php
```diff
@@ -70,4 +71,26 @@ public function getCloudFlareEmail()
...
+    /**
+     * @param $value
+     *
+     * @return bool
+     */
+    public function setIpRewrite($value)
+    {
+        // Clear option with temp value
+        update_option(self::IP_REWRITE, 'cleared');
```
jwineman: Why are you "clearing" the value here?

src/Wordpress/PluginActions.php
```diff
@@ -52,4 +52,65 @@ public function loginWordPress()
...
+
+    /**
+     * GET /plugin/:zonedId/settings.
+     *
+     * @return mixed
+     */
+    public function getPluginSettings()
+    {
+        // Always return true
+        $value = true;//$this->dataStore->getIpRewrite();
```
jwineman: Why does this always return `true`?  Shouldn't we always be returning `$this->dataStore->getIpRewrite();`

src/Wordpress/PluginActions.php
```diff
@@ -52,4 +52,65 @@ public function loginWordPress()
...
+                'value' => $value,
+                'editable' => true,
+                'modified_on' => '',
+                ),
+            )
+        );
+
+        $response['errors'] = []; // TODO: This doesn't seem a nice way
+        $response['messages'] = [];
+
```
jwineman: Correct - this isn't the ideal way.  Lets add these fields as part of `createAPISuccessResponse()` since every response will need empty arrays for errors and messages.

src/Wordpress/PluginActions.php
```diff
@@ -52,4 +52,65 @@ public function loginWordPress()
...
+        $value = true;//$this->dataStore->getIpRewrite();
+
+        $response = $this->api->createAPISuccessResponse(
+            array(
+                array(
+                'id' => 'ip_rewrite',
+                'value' => $value,
+                'editable' => true,
+                'modified_on' => '',
+                ),
```
jwineman: Since every individual setting will have the same structure (id, value, editable, modified on) lets create a method that takes in those four parameters as arguments and returns that array as the response.  Default editable to true if its not passed in and default modified on to empty string if its not passed in.  This will make adding new settings easy in the future.

src/Wordpress/DataStore.php
thellimist: This was a small hack. If there is a chance of setting the same value, `update_option` will return `false` which will return `createAPIError` in the API response. 

I removed this code.

Since the API call is toggle(for now) I believe no such thing will happen. If there is a possibility of setting the value intentional to the same value, the current code (not this one) will return `false`.

src/Wordpress/DataStore.php
```diff
@@ -70,4 +73,46 @@ public function getCloudFlareEmail()
...
+     */
+    public function getPluginSettings($api)
+    {
+        $ip_rewrite_value = get_option(self::CLOUDFLARE_SETTING_PREFIX + self::IP_REWRITE);
+        $protocol_rewrite_value = get_option(self::CLOUDFLARE_SETTING_PREFIX + self::PROTOCOL_REWRITE);
+
+        $settings = [];
+        array_push($settings, $api->createPluginResult(self::IP_REWRITE, $ip_rewrite_value, true, ''));
+        array_push($settings, $api->createPluginResult(self::PROTOCOL_REWRITE, $protocol_rewrite_value, true, ''));
+
```
jwineman: Yes!  This is exactly what I was thinking - nice job.  Very clean, easily extendible in the future.

PR #12, opened 2016-06-20
correct overview link

peterood:
https://www.cloudflare.com/overview.html returns "The page you are looking for cannot be found."


jwineman:
Hi Peter,

Great catch!  Appreciate the fix, thanks for your contribution!

Thanks,
John

PR #13, opened 2016-06-21
added php and wordpress version limit check

thellimist:


cloudflare.php
```diff
@@ -42,6 +42,30 @@
...
+    $min_php_version = '5.3';
+
+    if (version_compare(PHP_VERSION, $min_php_version, '<')) {
+        $flag = 'PHP';
+    } elseif (version_compare($wp_version, $min_wp_version, '<')) {
+        $flag = 'WordPress';
+    } else {
+        return;
+    }
+    $version = 'PHP' == $flag ? $min_php_version : $min_wp_version;
```
jwineman: I don't like this logic, its too confusing what about:
```
if(!version_compare(PHP_VERSION, $min_php_version, '<') || !version_compare($wp_version, $min_wp_version, '<')) {
//Deactivate Plugin

//Kill Execution
}
```

cloudflare.php
```diff
@@ -42,6 +42,30 @@
     }
 }
 
+// Call when the Plugin is activated in server. 
+function cloudflare_activate()
+{
+    global $wp_version;
+    $min_wp_version = '3.1';
+    $min_php_version = '5.3';
```
jwineman: move these into `const`s at the top of the file and add a trialing 0 to the end just to be explicit.

PR #19, opened 2016-07-13
Thellimist/pi 670

thellimist:


src/WordPress/PluginActions.php
```diff
@@ -27,6 +28,7 @@ public function __construct(DefaultIntegration $defaultIntegration, APIInterface
         $this->wordpressAPI = $defaultIntegration->getIntegrationAPI();
         $this->dataStore = $defaultIntegration->getDataStore();
         $this->logger = $defaultIntegration->getLogger();
+        $this->integration = $defaultIntegration;
```
jwineman: Why did you add this?  I don't see it being used anywhere?  If its not being used lets remove it.  If it is being used lets talk about it.

src/WordPress/PluginActions.php
```diff
@@ -60,7 +62,6 @@ public function loginWordPress()
      */
     public function getPluginSettings()
     {
-        // Always return true
         $settings = $this->dataStore->getPluginSettings($this->api);
```
jwineman: Don't pass `$this->api` to `CF\WordPress\DataStore.php`.  The DataStore is only for getting/setting data in WordPress, it shouldn't be interacting with the API at all.  It looks like you're using it to create API responses, just do that here instead since you already have access to `$this->api` here.

src/WordPress/PluginActions.php
jwineman: Instead of doing this lets extend the core functionality. 
1. Create `CF\WordPress\Router\WordPressPluginAPIRouter.php`
2. Extend `route(Request $request)` to pass in a `CF\WordPress\API\WordPressClientAPI`
3. Update this constructor to accept `CF\WordPress\API\WordPressClientAPI`
4. `$this->wordPressAPIClient = $wordPressAPIClient

src/WordPress/PluginRoutes.php
```diff
@@ -23,6 +23,16 @@ class PluginRoutes
             ),
         ),
 
+        // Must be above "plugin/:id/settings/:human_readable_id"
+        'plugin/:id/settings/default_settings' => array(
```
jwineman: Instead of hard coding this route just have `plugin/:id/settings/:human_readable_id` go to a method that switches on `:human_readable_id` and then returns the method that handles that ID.  This functionality working correctly as is depends on the order of these routes being correct which has a lot of potential for bugs.

PR #25, opened 2016-07-15
Thellimist/pi 682

thellimist:


src/WordPress/DataStore.php
```diff
@@ -120,18 +117,18 @@ public function setPluginSetting($settingId, $value)
             return false;
         }
 
-        return update_option(self::CLOUDFLARE_SETTING_PREFIX.$settingName, $value);
+        return update_option($settingName, $value);
     }
 
     private static function getPluginSettingName($settingId)
```
jwineman: If you move `$settings` from `getPluginSettings()` into a class variable you can just do:
```
return array_key_exists($settingId, $settings);
```

PR #27, opened 2016-07-19
Thellimist/pi 689

thellimist:


src/Test/Wordpress/PluginActionsTest.php
```diff
@@ -61,36 +61,72 @@ public function testLoginWordPressFail()
         $this->assertEquals(false, $response['success']);
     }
 
-    public function testPatchPluginSettings()
+    public function testGetPluginSettings()
```
jwineman: Still need a more descriptive test name.  Something like `testGetPluginSettingsHandlesValidSetting()`.  Need to know what the expected result of the test is.

src/Test/Wordpress/PluginActionsTest.php
```diff
@@ -61,36 +61,72 @@ public function testLoginWordPressFail()
...
+        ->setMethods(array('patchPluginDefaultSettings'))
+        ->getMock();
+
+        $pluginActions->expects($this->once())
+                    ->method('patchPluginDefaultSettings');
+
+        $pluginActions->patchPluginSettingsRouter();
+    }
+
+    public function testPatchPluginSettings()
```
jwineman: Need a more descriptive name here too.

PR #28, opened 2016-07-20
PI-691 fix WordPress AdminPanel settings css

thellimist:


jwineman:
This isn't related to the PR but can you pull in the CSS from cloudflare/cloudflare-frontend again please?  I removed the reset CSS for magento and now our styles are MUCH closer to WWW.  I don't think it will fix/break anything but it should bring the look and feel closer to that of WWW.

PR #32, opened 2016-07-27
Added SSL header correction functionality into plugin.

IcyApril:
Allows Flexible SSL to work transparently.

### DIFF
--- a/cloudflare.php
+++ b/cloudflare.php
@@ -95,8 +95,7 @@ function cloudflare_init()
     $ipRewrite = new IpRewrite();
     $is_cf = $ipRewrite->isCloudFlare();
     if ($is_cf) {
-        // The HTTP Request is from Cloudflare. Ip is rewritten successfully.
-        // For more info: github.com/cloudflare/cf-ip-rewrite
+        sslRewrite();
     }
 
     add_action('admin_menu', 'cloudflare_config_page');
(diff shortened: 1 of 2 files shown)


jwineman:
I approve.  @thellimist feel free to merge if you're okay.

thellimist:
@IcyApril 
also could you do 
`composer install`
`composer format`

Thank you

cloudflare.php
```diff
@@ -468,6 +469,16 @@ function purgeCache()
...
 }
 
+function sslRewrite() {
+    if ( isset( $_SERVER['HTTP_X_FORWARDED_PROTO'] ) && $_SERVER['HTTP_X_FORWARDED_PROTO'] == 'https' ) {
+        $_SERVER['HTTPS'] = 'on';
+
+        return true;
+    }
+
+    return false;
```
jwineman: They're useful if we ever wrap the call in an if statement
```
if(sslRewrite()) {
//do something if we rewrote the header
}
```

jwineman:
```
vendor/bin/phpcbf -n --standard=PSR2 --extensions=php,live.php src/
```
per https://github.com/squizlabs/PHP_CodeSniffer/wiki/Fixing-Errors-Automatically#using-the-php-code-beautifier-and-fixer

jwineman:
basically `phpcbf` instead of `phpcs` for whatever command you run.

cloudflare.php
jwineman: Also if we were to test `sslRewrite()` it would need to return `true` or `false` so we could `assert` whether it exited correctly.

thellimist: We can not write tests for sslRewrite if it's in `cloudflare.php`. It'll need to be moved inside `src/WordPress`.

jwineman: I know.  We'll figure that out at some point.  I only mention it to point out that we should be writing all our code in a way that is easy to test even if we can't actually write a test for it.

PR #33, opened 2016-08-02
Thellimist/pi 717

thellimist:


src/WordPress/DataStore.php
```diff
@@ -68,7 +68,7 @@ public function getClientV4APIKey()
      */
     public function getHostAPIUserKey()
     {
-        return null;
+        return;
```
jwineman: why did you remove the `null` here?

thellimist: commit c554533 is done automatically by php code sniffer. I set sublime to run `phpcs -n --standard=PSR2` in every save for php files. 

https://github.com/php-fig-rectified/fig-rectified-standards/blob/master/PSR-2-R-coding-style-guide-additions.md#return-void-vs-null

PR #34, opened 2016-08-03
Thellimist/pi 715

thellimist:


src/WordPress/WordPressClientAPI.php
```diff
@@ -15,10 +15,10 @@ class WordPressClientAPI extends Client
     public function getZoneTag($zone_name)
     {
         $request = new Request('GET', 'zones/', array('name' => $zone_name), array());
-        $response = self::callAPI($request);
+        $response = $this->callAPI($request);
```
jwineman: Don't bother submitting a PR to fix it but in the future you want to use `$this->` to when calling non static members and `self::` when referring to static ones. https://stackoverflow.com/a/151976

thellimist: That's why I changed from `self::` to `$this->`.  Since `public function callAPI(Request $request)` is not static `$this->` should be correct if I'm not missing something.

jwineman: Oh my bad, this is what I get for doing code reviews late at night.  Yeah you changed it correctly.

PR #36, opened 2016-08-08
Thellimist/pi 709

thellimist:
First 3 commits are not related with the PR. They are commits which didn't belong to any ticket.


src/Test/Wordpress/WordPressAPITest.php
```diff
@@ -1,34 +0,0 @@
-<?php
```
jwineman: Why did we remove this?

thellimist: Initially, I removed it and deleted the test cause we didn't need it. But I reverted the commit. I'll add the tests back again.

PR #39, opened 2016-08-11
PI-718 added subdomain check

thellimist:
@jwineman Don't merge since the requirements are not given yet the PR could change


cloudflare.php
```diff
@@ -502,3 +502,7 @@ function theme_save_pressed()
 // Enable HTTP2 Server Push
 require_once plugin_dir_path(__FILE__).'src/HTTP2ServerPush.php';
 add_action('init', array('HTTP2ServerPush', 'init'));
+
+// Display error message if user is using a subdomain
+require_once plugin_dir_path(__FILE__).'src/SubdomainChecker.php';
```
jwineman: Is there are reason we aren't using composer's here?
```
require_once(plugin_dir_path(__FILE__) .'vendor/autoload.php');
```

jwineman:
How come you didn't implement this as a hook around the GET /zones call? 
https://github.com/cloudflare/CloudFlare-Wordpress/blob/master/src/WordPress/ClientActions.php#L38

What you could do is loop through the results of GET /zones in `returnWordPressDomain()`, if the root domain is a substring of the domain in `CF\WordPress\WordPressClientAPI::getDomainList()` then you know you're dealing with a subdomain and you can return that domain. 

Then on cloudflare/cloudflare-frontend you can just check the domain returned by the backend (which will always be a root domain now) against the current URL if have a react component that displays conditionally if they're not equal.

src/SubdomainChecker.php
```diff
@@ -0,0 +1,67 @@
...
+    private static $isSubdomain = false;
+    private static $originalDomain = '';
+    private static $subDomain = '';
+
+    public static function init()
+    {
+        if (!self::$initiated) {
+            self::init_hooks();
+        }
+
```
jwineman: Why did you choose to make this class static? Isn't your `init()` method basically a constructor?  Why not put all that logic and the `new CF\*` classes in a constructor? 

If you stick with this approach make sure to write tests for all these methods too.

jwineman: Is this something all plugins can use?  Should we move it to cloudflare/cloudflare-plugin-backend?

cloudflare.php
thellimist: If I understood your question correctly the answer is yes. We are using 
```
add_action('admin_notices', array('SubdomainChecker', 'enableSubdomainNotice'));
```
hook. Hooks and requests should have different structure. We added every WP hook to `cloudflare.php` which was a mess. I wanted to simplify `cloudflare.php` file and separate every functionality which uses hooks to different classes. I figured I can create `static class`es only for WP hooks. Currently I'm adding hooks to a single file under `Cloudflare-WordPress\src\`. To improve the structure we can add `Cloudflare-WordPress\src\WordPress\Hooks` which is on my todo list.

thellimist:
> How come you didn't implement this as a hook around the GET /zones call? 
> https://github.com/cloudflare/CloudFlare-Wordpress/blob/master/src/WordPress/ClientActions.php#L38


That wouldn't work. We've hooks which rely on `WordPressAPI::getDomainList`. Since hooks can be called anytime we needed to have a global domain checker which checks if the domain is subdomain even before a single request is made. 

For example [purgeCache](https://github.com/cloudflare/CloudFlare-Wordpress/blob/master/cloudflare.php#L448-L474) functions works in a hook which is activated when you change the WordPress template and it requires `WordPressAPI::getDomainList` function. This is a WordPress issue.

> What you could do is loop through the results of GET /zones in returnWordPressDomain(), if the root domain is a substring of the domain in CF\WordPress\WordPressClientAPI::getDomainList() then you know you're dealing with a subdomain and you can return that domain.
> 
> Then on cloudflare/cloudflare-frontend you can just check the domain returned by the backend (which will always be a root domain now) against the current URL if have a react component that displays conditionally if they're not equal.

On the other hand we can do the solution you suggested and have another logic in the frontend to handle subdomains info messages which we can have in all integrations.

src/SubdomainChecker.php
thellimist: > Why did you choose to make this class static? Isn't your init() method basically a constructor? Why not put all that logic and the new CF\* classes in a constructor?
> 
> If you stick with this approach make sure to write tests for all these methods too.

We can do that but since I used a hook I wanted to go with the style I used with [ServerPush](https://github.com/cloudflare/CloudFlare-Wordpress/blob/master/src/HTTP2ServerPush.php). But if we're going to remove the logic of the error message to frontend instead of backend I can remove the `admin_notices` hook and do as you suggested.

> Is this something all plugins can use? Should we move it to cloudflare/cloudflare-plugin-backend?

Init CF\* classes in a constructor? Probably but I wouldn't add this to `cloudflare/cloudflare-plugin-backend` until we're certain this is a good solution.

jwineman:
I'll leave it up to you to decide.  I had envisioned it as separate logic on the frontend but either solution is fine.

cloudflare.php
jwineman: What I meant was Composer's auto loader should include everything in src/ so you should be able to do `vendor/autoload.php` at the top of the file and then remove the requires for SubdomainChecker.php and HTTP2ServerPush.php

src/WordPress/WordPressClientAPI.php
```diff
@@ -31,6 +31,49 @@ public function getZoneTag($zone_name)
...
+     *
+     * @param $haystack 
+     * @param $needle 
+     * 
+     * @return bool
+     */
+    private function endsWith($haystack, $needle)
+    {
+        // search forward starting from end minus needle length characters
+        return $needle === '' || (($temp = strlen($haystack) - strlen($needle)) >= 0 && strpos($haystack, $needle, $temp) !== false);
```
terinjokes: Why not chop the `strlen($needle)`  off the end of `$haystack` and compare?

```php
private function endsWith($haystack, $needle)
{
    $len = strlen($needle);

    // chop string starting at `$len` characters from the end,
    //     continue for length `$len` characters,
    // then compare against `$needle`
    return $needle === strpos($haystack, -$len, $len);
}
```

src/WordPress/SubdomainChecker.php
```diff
@@ -0,0 +1,39 @@
...
+        $config = new \CF\Integration\DefaultConfig('[]');
+        $logger = new \CF\Integration\DefaultLogger($config->getValue('debug'));
+        $dataStore = new \CF\WordPress\DataStore($logger);
+        $wordpressAPI = new \CF\WordPress\WordPressAPI($dataStore);
+        $wordpressIntegration = new \CF\Integration\DefaultIntegration($config, $wordpressAPI, $dataStore, $logger);
+        $clientAPIClient = new \CF\WordPress\WordPressClientAPI($wordpressIntegration);
+
+        $wpDomain = $wordpressAPI->getOriginalDomain();
+
+        if ($clientAPIClient->isSubdomain($wpDomain)) {
```
jwineman: `$clientAPIClient` doesn't have a `isSubdomain()` method.  The `WordPressClientAPI` should only contain logic around interacting with the Client V4 API.  If you did forget to commit this method lets move it to the `WordPressAPI` instead or move it into this class.

src/WordPress/SubdomainChecker.php
```diff
@@ -0,0 +1,39 @@
...
+    private $subDomain = '';
+
+    public function changeDomainName()
+    {
+        $config = new \CF\Integration\DefaultConfig('[]');
+        $logger = new \CF\Integration\DefaultLogger($config->getValue('debug'));
+        $dataStore = new \CF\WordPress\DataStore($logger);
+        $wordpressAPI = new \CF\WordPress\WordPressAPI($dataStore);
+        $wordpressIntegration = new \CF\Integration\DefaultIntegration($config, $wordpressAPI, $dataStore, $logger);
+        $clientAPIClient = new \CF\WordPress\WordPressClientAPI($wordpressIntegration);
```
jwineman: Hard coding your dependencies like this makes `changeDomainName()` hard to test because you can't mock any of the objects you instantiate with `new`.  This is why we use constructor based dependency injection.  

If this method is only used in `WordPressAPI::domainName` why not move it into that class?  

If you decide to keep this as a class, move the dependencies into the constructor so we can test them, otherwise just move it into `WordPressAPI`.  Either way we need tests for this method too.

src/WordPress/WordPressAPI.php
```diff
@@ -8,6 +8,7 @@
 class WordPressAPI implements IntegrationAPIInterface
 {
     private $dataStore;
+    private static $domainName;
```
jwineman: Why is this static?  This class has a constructor and isn't intended to be used statically.

thellimist: `static $domainName` would not be garbage collected when a regular request is made. When an action which calls a hook would need to use the `getDomainList`, `domainName` would still be set and we won't have to make another API call.
```
    public function getDomainList($userId = null)
    {
        return array(self::$domainName);
    }
```

src/WordPress/SubdomainChecker.php
thellimist: clientAPIClient does have [isSubdomain](https://github.com/cloudflare/CloudFlare-Wordpress/blob/thellimist/PI-718/src/WordPress/WordPressClientAPI.php#L57-L73). The reason I put there was to avoid another set of objects to be created inside `WordPressAPI` cause I am using `GET /zones` [client](https://github.com/cloudflare/CloudFlare-Wordpress/blob/thellimist/PI-718/src/WordPress/WordPressClientAPI.php#L62) call. 

If you still think it belongs inside `WordPressAPI` I can move it.

src/Test/WordPress/WordPressAPITest.php
```diff
@@ -46,14 +58,46 @@ public function invokeMethod(&$object, $methodName, array $parameters = array())
...
+        $domainName = 'domain.com';
+
+        $this->mockDataStore = $this->getMockBuilder('CF\WordPress\DataStore')
+                ->disableOriginalConstructor()
+                ->setMethods(array('getDomainNameCache'))
+                ->getMock();
+
+        $this->mockDataStore->method('getDomainNameCache')->willReturn($domainName);
+
+        $this->wordpressAPI = new WordPressAPI($this->mockDataStore);
```
jwineman: Should move this into `setup()`.  Any changes you make to the `$mockDataStore` will still apply even after you've instantiated `$wordpressAPI`.

src/WordPress/WordPressAPI.php
```diff
@@ -69,6 +70,48 @@ public function getHostAPIKey()
...
+     * other plugins.
+     * 
+     * @param null $userId
+     *
+     * @return mixed
+     */
+    public function getDomainList($userId = null)
+    {
+        $cachedDomainName = $this->dataStore->getDomainNameCache();
+        if (empty($cachedDomainName)) {
```
jwineman: Shouldn't the logic be:

1. Check cache for value.  
2. If empty set cache equal to output of `getOriginalDomain()`
3. return domain.

If for some reason the cache fails this method will return an empty array which is incorrect behavior since this method should always return a MINIMUM of one result.

src/WordPress/DataStore.php
```diff
@@ -65,6 +66,27 @@ public function getHostAPIUserKey()
     }
 
     /**
+     * @return mixed
+     */
+    public function getDomainNameCache()
```
jwineman: You're duplicating some of your cache logic.  Why not do it all in WordPressAPI::getDomainList()? 
```
//CF\WordPress\WordPressAPI.php
    public function getDomainList($userId = null)
    {
        $cachedDomainName = $this->get(DataStore::CACHED_DOMAIN_NAME);
        if (empty($cachedDomainName)) {
             $cachedDomainName = $this->get(self::CACHED_DOMAIN_NAME);
             $this->set(DataStore::CACHED_DOMAIN_NAME, $cachedDomainName);
        }
        return array($cachedDomainName);
    }
```

src/WordPress/WordPressAPI.php
```diff
@@ -86,23 +129,40 @@ private function formatDomain($domainName)
         return $formattedDomain;
     }
 
+    public function setWordPressClientAPI($wordPressClientAPI)
```
jwineman: Why not pass this in as a constructor dependency?

src/WordPress/WordPressAPI.php
```diff
@@ -86,23 +129,40 @@ private function formatDomain($domainName)
...
 
-        return array($this->formatDomain($domainName));
+        return false;
     }
 
-    /**
-     * @return mixed
-     */
-    public function getUserId()
+    public function cacheWordPressDomainName($wordpressIntegration)
```
jwineman: Do you need the whole `$wordpressIntegration`?  Why not take specifically what you need and pass it in through the class's constructor?

src/WordPress/WordPressAPI.php
thellimist: The logic is [here](https://github.com/cloudflare/CloudFlare-Wordpress/blob/f0bda21af307943e7da0f727eceed7fdea5d4e38/src/WordPress/WordPressAPI.php#L155-L168). 

1. If the cached value and the original domain are not equal
2. Cache it

so even if `$cachedDomainName` in `getDomainList` returns empty the cache will be set to `getOriginalDomain()`. I wanted `DataStore` to only fetch and pass data so I put the logic in `WordPressAPI`.

The reason I added `if (empty($cachedDomainName))` is I wanted to make [this](https://github.com/cloudflare/CloudFlare-Wordpress/blob/f0bda21af307943e7da0f727eceed7fdea5d4e38/src/WordPress/WordPressAPI.php#L159) check.

PR #40, opened 2016-08-17
Thellimist/pi 736

thellimist:


src/Test/WordPress/PluginActionsTest.php
```diff
@@ -0,0 +1,183 @@
...
+                )
+            );
+            $this->mockWordPressClientAPI->method('responseOk')->willReturn(true);
+            $this->mockWordPressClientAPI->method('changeZoneSettings')->willReturn(true);
+            $this->mockWordPressClientAPI->method('createPageRule')->willReturn(true);
+
+            $this->mockWordPressClientAPI->expects($this->exactly(15))->method('changeZoneSettings');
+            $this->mockWordPressClientAPI->expects($this->exactly(2))->method('createPageRule');
+
+            $pluginActions = $this->getMockBuilder('CF\WordPress\PluginActions')
```
jwineman: This doesn't need to block you from merging but every test builds this `$pluginActions` mock why not move it into `setup()`?

src/WordPress/PluginActions.php
```diff
@@ -6,139 +6,39 @@
...
+     * PATCH /plugin/:id/settings/default_settings               
      *
-     * @return mixed
+     * Requests are syncronized
      */
-    public function getPluginSettings()
+    public function applyDefaultSettings()
     {
-        $settings = $this->dataStore->getPluginSettings();
+        $this->wordPressClientAPI = $this->createWordPressClientAPI($this->defaultIntegration);
```
jwineman: Why wrap this in a method since `createWordPressClientAPI()` just returns it with no extra logic?

src/WordPress/PluginRoutes.php
```diff
@@ -2,34 +2,20 @@
...
+    /**
+     * @param $routeList
+     *
+     * @return mixed
+     */
+    public static function getRoutes($routeList)
+    {
+        foreach ($routeList as $routePath => $route) {
+            $route['class'] = '\CF\WordPress\PluginActions';
+            $routeList[$routePath] = $route;
```
jwineman: Is this line necessary?  Won't `$routeList[$routePath]` already be equal to `$route` even if you don't set it here?

thellimist: I just tested and it doesn't recognize the routes without that line. But you do have a point it shouldn't be needed 😕

src/WordPress/PluginActions.php
thellimist: I needed to mock `$this->wordPressClientAPI` so I added a wrapper

src/WordPress/PluginRoutes.php
jwineman: Ah nevermind, its cause `$route` is scoped locally to the function.  This is fine.

src/WordPress/PluginActions.php
jwineman: Ah okay thats fine I guess.  Typically you would do this with a setter:
```
public function setWordPressClientAPI(WordPressClientAPI $wordpressClientAPI) {
   $this->wordpressClientAPI = $wordprssClientAPI;
}
```
and called that method with your mock in the `setup()` of the test.

jwineman: https://github.com/cloudflare/cloudflare-plugin-backend/blob/master/src/API/AbstractPluginActions.php#L35-L43

PR #42, opened 2016-08-18
PI-745: Removing dead code from cloudflare.php.

jwineman:


cloudflare.php
```diff
@@ -411,40 +182,6 @@ function cloudflare_buffer_wrapup($buffer)
...
-    ob_start('cloudflare_buffer_wrapup');
-}
-// Protocol Rewrite Hook # 1
-// ob_start (buffer) is also being used with HTTP2ServerPush. There can
-// be conflicts between the buffers when protocol rewrite is reactivated.
-// add_action('plugins_loaded', 'cloudflare_buffer_init');
-
-// wordpress 4.4 srcset ssl fix
-// Shoutout to @bhubbard: https://wordpress.org/support/topic/44-https-rewritte-aint-working-with-images?replies=12
-function cloudflare_ssl_srcset($sources)
```
jwineman: Not doing protocol rewrite anymore, don't need this right?

cloudflare.php
```diff
@@ -411,40 +182,6 @@ function cloudflare_buffer_wrapup($buffer)
     return $buffer;
 }
 
-function cloudflare_buffer_init()
```
jwineman: HTTP2 ServerPush does its own buffer init so we dont need this right?

cloudflare.php
```diff
@@ -8,43 +8,11 @@
...
-
-$cfPostKeys = array('cloudflare_zone_name', 'cf_key', 'cf_email', 'dev_mode', 'protocol_rewrite');
-
 const MIN_PHP_VERSION = '5.3';
 const MIN_WP_VERSION = '3.4';
 
-foreach ($_POST as $key => $value) {
-    if (in_array($key, $cfPostKeys)) {
-        $_POST[$key] = cloudflare_filter_xss($_POST[$key]);
-    }
```
thellimist: Why did we remove this?

cloudflare.php
```diff
@@ -79,17 +47,6 @@ function set_default_keys()
...
-function cloudflare_filter_xss($input)
-{
-    return htmlentities($input, ENT_QUOTES | ENT_HTML5, 'UTF-8');
-}
-
-// Make sure we don't expose any info if called directly
-if (!function_exists('add_action')) {
-    echo "Hi there!  I'm just a plugin, not much I can do when called directly.";
-    exit;
-}
```
thellimist: Currently this is useless since the user can see other files by url like
mydomain.com/wordpress/wp-content/plugins/cloudflare/src/anyPHPFiles.php

I'll need to find a general solution to disable the user to look at any one of them. 

I don't think it's a vulnerability since the code is already on github but...

cloudflare.php
thellimist: This is related with protocol rewrite functionality. If we are removing protocol rewrite this is useless.

cloudflare.php
```diff
@@ -316,68 +148,7 @@ function match_domain_to_zone($domain, $zones)
     return;
```
thellimist: This is also useless. I just realized this was here. I could have used it in subdomain check code 😄 It's also buggy. It doesn't work with `domain.co.uk` 😄

cloudflare.php
```diff
@@ -411,40 +182,6 @@ function cloudflare_buffer_wrapup($buffer)
     return $buffer;
```
thellimist: If we are removing protocol rewrite this should also be removed.

cloudflare.php
jwineman: It only filters `$cfPostKeys` which are only used in the old plugin.

cloudflare.php
```diff
@@ -8,43 +8,11 @@
...
-
-*/
-
 require_once 'vendor/autoload.php';
 
-define('CLOUDFLARE_VERSION', '1.3.24');
-define('CLOUDFLARE_API_URL', 'https://www.cloudflare.com/api_json.html');
-define('CLOUDFLARE_SPAM_URL', 'https://www.cloudflare.com/ajax/external-event.html');
-
-use \CloudFlare\IpRewrite;
```
thellimist: We use IpRewrite

PR #48, opened 2016-09-01
PI-744: Input sanitization for CF\WordPress\DataStore.

jwineman:


ejcx:
LGTM

thellimist:
https://core.trac.wordpress.org/browser/tags/4.5.3/src/wp-includes/option.php#L245

> 245	 * @param string      $option   Option name. Expected to not be SQL-escaped.
> 246	 * @param mixed       $value    Option value. Must be serializable if non-scalar. Expected to not be SQL-escaped.

I don't think we should escape it. WordPress Options API looks safe

jwineman:
Hmmm.  You're right.  I'll follow up with the person who suggested this and make sure we're covering all our bases.

PR #51, opened 2016-09-07
Thellimist/pi 734 updated cloudflare-plugin-backend and cloudflare-frontend 

thellimist:
Upgraded backend 1.1.7
Upgraded frontend 2.3.1
Fix changes including backend problems


src/WordPress/DataStore.php
```diff
@@ -18,6 +20,13 @@ class DataStore implements DataStoreInterface
     public function __construct(DefaultLogger $logger)
     {
         $this->logger = $logger;
+
+        // Create temp object to be able to call createPluginSettingObject
+        $config = new DefaultConfig('[]');
+        $wordpressAPI = new WordPressAPI($this);
+        $wordpressIntegration = new DefaultIntegration($config, $wordpressAPI, $this, $logger);
+
+        $this->tempPlugin = new Plugin($wordpressIntegration);
```
thellimist: Every solution I came up with sucked which included using the current version of cloudflare-plugin-backend. For fast fix I chose this one.

thellimist:
@jwineman 
I need to try the other solution we mentioned. You don't have to review.

PR #54, opened 2016-09-09
Jwineman/pi 792 wordpress ajax

jwineman:


index.php
```diff
@@ -1,2 +1,60 @@
 <?php
-# Silence is golden.
+if (!defined('ABSPATH')) { // Exit if accessed directly
+	exit;
+}
+require_once 'vendor/autoload.php';
```
thellimist: wasn't the whole idea is to get rid of `require_once 'vendor/autoload.php';`

jwineman: I haven't tested, we should be able to now.  All I wanted in the PR was to switch to WordPress AJAX and get rid of `wp-load.php`.  With this change done we should be able to remove `vendor/autoload.php` and move ALL class instantiations to cloudflare.php.

PR #70, opened 2016-09-21
PI-823: Fixing uninstall hook.

jwineman:
I'm not sure if you can uninstall a plugin without deactivating it first.  If thats the case can't we just delete this hook since `deactivate()` calls `clearDataStore()` as well?

### DIFF
--- a/cloudflare.php
+++ b/cloudflare.php
@@ -16,7 +16,7 @@ if (!defined('ABSPATH')) { // Exit if accessed directly
 
 // ************************************************************** //
 
-// Initialize Global Objects 
+// Initialize Global Objects
 $cloudflareConfig = new CF\Integration\DefaultConfig(file_get_contents('config.js', true));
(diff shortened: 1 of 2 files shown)


thellimist:
https://github.com/cloudflare/CloudFlare-Wordpress/blob/a69c346dbd6ea008f7be5753d4f30badb3f16853/src/WordPress/Uninstall.php

this should be removed

thellimist:
WordPress doesn't call deactivate when uninstall hook is trigger so we need to handle it in both cases.

jwineman:
Yeah I know it doesn't call deactivate when you uninstall.  You can't call uninstall UNTIL you've already deactivated right?

thellimist:
Yes.

thellimist:
Removed `Uninstall.php` file

https://github.com/cloudflare/CloudFlare-Wordpress/commit/839becfedc00f651595f2c41b925801b09695b7b

PR #75, opened 2016-09-21
Losslessly compress images

XhmikosR:
Total difference;

```
Before: 122.3 KB (125.410 bytes)
After:  102.9 KB (105.550 bytes)
```


jwineman:
I need to test this on my development environment and run it by our design team but assuming there are no issues I'll merge early next week probably.

XhmikosR:
@jwineman: shouldn't this change be in https://github.com/cloudflare/CloudFlare-FrontEnd?

I didn't know at the time that you had that repo and this one depended on that.

PR #76, opened 2016-09-21
Normalize whitespace.

XhmikosR:
Use 4 spaces consistently. Also, change .editorconfig to reflect this.

See https://github.com/cloudflare/CloudFlare-Wordpress/pull/76/files?w=1 for the non-whitespace diff.


jwineman:
Our editor config is based off the WordPress coding standards:
https://make.wordpress.org/core/handbook/best-practices/coding-standards/

They use tabs which is why we use tabs.
https://core.trac.wordpress.org/attachment/ticket/26429/.editorconfig.patch

XhmikosR:
But you don't use tabs consistently either.

jwineman:
I know.  We try to format to [PSR-2](http://www.php-fig.org/psr/psr-2/) but WordPress doesn't conform to PSR-2 so it sorta messes everything up.

XhmikosR:
Idiotic standard if you ask me let alone they still use SVN.

Oh well, just thought I'd make things consistent and trim trailing spaces
too along the way.

jwineman:
Agreed.  We're debating internally whether to drop support for their .editorconfig and just move back to PSR-2.  We were trying to do things "the wordpress way" as much as possible but sometimes it doesn't seem worth it.

3.0.0 of this plugin had a script that would deploy from git to their SVN too but it was buggy so now we're back to doing manual SVN deploys. 

But yeah we rely heavily on composer and PHP namespaces so thats why we now only support PHP 5.3.10+.  The main benefit of this is we can avoid all the WordPress best practices that exist to help you avoid all the global namespace conflicts.

PR #78, opened 2016-09-21
PI-825: Moved PHP version check to cloudflare.php

jwineman:


src/WordPress/Hooks.php
```diff
@@ -82,47 +79,17 @@ public function initProxy()
 
     public function activate()
     {
-        $this->checkVersionCompatibility();
-        $this->checkDependenciesExist();
```
thellimist: Why did we remove the functions? Functions provide self-documentation and make the code more readable.

cloudflare.php
```diff
@@ -14,6 +14,15 @@
     exit;
 }
 
+const CF_MIN_PHP_VERSION = '5.3.10';
+const CF_MIN_WP_VERSION = '3.4';
```
thellimist: `const` keyword is supported PHP 5.3+ 

If the user has a lower version it wouldn't matter cause It'll just give a fatal error. We should change it to a normal variable.

thellimist: Why is `MIN_PHP_VERSION` `5.3.10`. Is there a reason it's not `5.3`?

jwineman: Its changed to 5.3.10 because the root cause of this bug is a user running 5.3.3, and the issue wasn't fixed until 5.3.9

src/WordPress/Hooks.php
jwineman: I think its fine.  `activate()` documents the function gets called during activation and the `wp_die()` errors tell you its version related.  I was trying to avoid functions in Hooks.php that aren't directly mapped to an actual WordPress hook.

cloudflare.php
jwineman: I'll update the const to a normal variable.

src/WordPress/Hooks.php
thellimist: I don't agree with `avoid functions in Hooks.php `. As long as the code is readable we can have many functions. Functions provide code documentation by default. 

In my opinion the code got relatively harder to read compared to the previous version still it's in a reasonable range to understand the code.

jwineman: Okay fair enough.  I won't try to refactor this kind of thing in the future.

PR #80, opened 2016-09-22
Jwineman/pi 827 global objects

jwineman:
We don't need to worry about the `new` keyword failing in early version of PHP because WordPress uses it in base code.


cloudflare.php
```diff
@@ -13,7 +13,6 @@
     exit;
 }
 
-define('CLOUDFLARE_MIN_WP_VERSION', '3.4');
```
thellimist: It's actually better to have it in cloudflare.php compared to Hooks since cloudflare.php is the first file people check and they can see which version is supported immediately.

src/WordPress/Hooks.php
```diff
@@ -98,7 +105,7 @@ public function activate()
             wp_die('<p><strong>Cloudflare</strong> plugin requires php5-curl to be installed.</p>', 'Plugin Activation Error', array('response' => 200, 'back_link' => true));
         }
 
-        return;
+        return true;
```
thellimist: we don't need this

src/WordPress/Hooks.php
```diff
@@ -12,6 +12,7 @@ class Hooks
     protected $api;
     protected $config;
     protected $dataStore;
+	protected $integrationContext;
```
thellimist: can you switch your editors config from tabs to spaces. Since everything is already space let's continue doing spaces.

cloudflare.php
thellimist: https://plugins.svn.wordpress.org/akismet/trunk/akismet.php

This is a good example of first file.

src/WordPress/Hooks.php
jwineman: The test for this method does `$this->assertTrue()`

src/WordPress/Hooks.php
jwineman: Isn't it the .editorconfig specifying tabs?  I'm down to just go back to PSR-2.

cloudflare.php
jwineman: I moved it to Hooks cause thats where the version check happened.  I don't care where the constant lives though.  Looks like you moved it back in the other PR, thats fine.

PR #88, opened 2016-09-27
Thellimist/pi 815 added tests

thellimist:


src/Test/WordPress/ProxyTest.php
```diff
@@ -2,78 +2,143 @@
...
+
+        $mockFileGetContents = $this->getFunctionMock('CF\WordPress', 'file_get_contents');
+        $mockFileGetContents->expects($this->any())->willReturn($jsonBody);
+
+        $request = $this->proxy->createRequest();
+
+        $this->assertFalse(isset($request->getParameters()['proxyURL']));
+        $this->assertFalse(isset($request->getParameters()['proxyURLType']));
+        $this->assertFalse(isset($request->getBody()['proxyURL']));
+        // $this->assertEquals($proxyURL, $request->getUrl());
```
thellimist: I commented this out cause `file_get_contents` is not getting mocked so the result is wrong.

If we change the `any()` in `$mockFileGetContents->expects($this->any())->willReturn($jsonBody);` to `once()` the test fails with 0 calls to `file_get_contents` mocked functions whereas it should be called once.

Our other tests in Datastore etc. all of them fail when tried to mock a global function with `$this->expects($number)`. 

The old version was not failing with the global scoped mocks.

jwineman: Rather than trying to mock `file_get_contents()` we can just do this:

1. ```protected $json;```
2.  In the constructor do ```$this->json = json_decode(file_get_contents('php://input'), true);```
3. Create ```public function setJson($json) { $this->json = $json; }
4. Update ```$body = $this->json```
5. In your test mock the response body with ```$this->proxy->setJson($mockJson);```

thellimist: @jwineman That'd work but the other tests are also not working (Datastore etc.)

jwineman: How are they breaking?  Do you need help coming up with a fix?

thellimist: They're not breaking. They're giving false positives. So basically they don't test but they trick us by acting like they test.

`$this->mockUpdateOption->expects($this->any())->willReturn(true);` this is a wrong assumption from the beginning. It should be `$this->mockUpdateOption->expects($this->once())->willReturn(true);` but the problem is even if you expect it `once` it'll always return 0 cause the initialization 

```
$this->mockUpdateOption = $this->getFunctionMock('CF\WordPress', 'update_option'); 
```

doesn't map with the original `update_option` function

thellimist: Instead of the method you suggested with `protected $jsonBody` and set the json in the contructor I took this approach for `file_get_content` because if we map the getting input logic to constructor then it'd have restricted our ability to move the constructor. 

Also I'm not even certain if it'd have worked in the current situation because we need to get the input when the hook calls `$this->proxy->run();` not before.

jwineman: Your approach is fine, I just don't like partial mocks.  

The constructor approach would have worked, when the class was instantiated it would grab the JSON, then it would be available when `run()` was called.  I'm not sure what you mean by restricting your ability to move the constructor.

thellimist: @jwineman 

> file_get_contents(), like basically all functions in PHP, is a blocking call.

That's why I only wanted it to be in run where it's guaranteed to have a request incoming.

PR #95, opened 2016-10-03
Thellimist/pi 847 cacheDomain moved

thellimist:


src/WordPress/ClientActions.php
```diff
@@ -14,6 +14,7 @@ class ClientActions
     private $dataStore;
     private $logger;
     private $request;
+    private $wordpressClientAPI;
```
jwineman: `$api` is `$wordpressClientAPI` for ClientActions right? I think you can remove `$wordpressClientAPI` and just use `$api`.

src/WordPress/ClientActions.php
```diff
@@ -64,4 +81,33 @@ public function returnWordPressDomain()
...
+    private function cacheDomainName($response)
+    {
+        // Check if domain name needs to cached
+        $wpDomain = $this->wordpressAPI->getOriginalDomain();
+        $cachedDomainList = $this->wordpressAPI->getDomainList();
+        $cachedDomain = $cachedDomainList[0];
+
+        if (Utils::getRegistrableDomain($wpDomain) !== $cachedDomain) {
+            // Since we may not be logged in yet we need to check the credentials being set
+            if ($this->wordpressClientAPI->isCrendetialsSet()) {
```
jwineman: Since this is part of the hook for GET /zones now this will never fire if they aren't logged in anymore right?  I think we can remove this check and `isCredentialsSet()`.

src/WordPress/ClientActions.php
```diff
@@ -64,4 +81,33 @@ public function returnWordPressDomain()
 
         return $cf_zones_list;
     }
+
+    private function cacheDomainName($response)
```
jwineman: Lets write tests for this method.

- Sets domain if not in cache
- Updates cache if domain has changed.

src/WordPress/ClientActions.php
thellimist: true, nice catch

src/WordPress/WordPressClientAPI.php
```diff
@@ -31,17 +31,6 @@ public function getZoneTag($zone_name)
     }
```
thellimist: @jwineman Removed

PR #96, opened 2016-10-04
PI-806: updated stylesheets/components.css to fix bug in active zone …

jwineman:
…selector which was hiding the options.


jwineman:
@thellimist FYI

PR #103, opened 2016-10-06
Clear Posts from Cloudflare cache

piscis:
Hi, thanks for the excellent plugin!

We're using Cloudflare to create full page cache (html + assets) of our site. Therefor we need to clear the cache when ever a post gets published.

With this pull request we added the ability to clear the cache by urls. When ever a post gets published we clear the homepage + the corresponding post (via permalink) from our cloudflare cache.

We also would like to configure the behaviour a bit more by extending the plugins settings interface, but could not figure out how to get the development environment for the repository **CloudFlare-FrontEnd** running. 

Could you help?

Thanks,

Alex


jwineman:
Hi,

Thanks for the PR!  Let me talk with our product manager and get back to you on this.  I know purge by URL was on our roadmap so we can probably use this.

As for cloudflare/cloudflare-frontend:

1. `npm install`
2. `gulp js`

The easiest way to "run" it is to symlink the compiled.js output from `gulp js` to the compiled.js for your wordpress blog.  What types of configuration were you wanting to make to the frontend?

piscis:
hi,

thanks for the info. There are basically three options we would like to control via wordpress:

1. To control if the cache for a post gets cleared, eg. When a post changes its status from draft to published 
2. A option to control if the homepage gets purged from cache when a post gets published
3. A option to control if corresponding tag and category pages getting purged if a post gets published.

At the moment 1. and 2. are hardcoded in the PR and very much tied down to our use case but I think it would be nice to have a option to control this behaviour.

As for option 3. this is a nice to have feature but could be useful.

Thanks for the info regarding the cloudflare-frontend.

jwineman:
Okay this works out great - I got the go ahead to merge this.  I just need to set aside some time to do some rework after merging for stuff that isn't your problem.  We're going to move it behind the "Automatic Cache Purge" toggle. 

As for the front end we probably can't accept any PRs you make there because that front end is shared across our [cPanel](https://github.com/cloudflare/CloudFlare-CPanel) and [Magento](https://github.com/cloudflare/CloudFlare-Magento) integrations as well.  The code that goes in there needs to be generic enough to fit the use cases for all three integrations, which is why we're moving it behind the Automatic Cache Purge toggle.  

However if you want to fork the front end and do something custom I'm more than happy to help you trouble shoot etc.

jwineman:
Hey,

We decided to merge https://github.com/cloudflare/Cloudflare-WordPress/pull/112 instead because it was going to require less rework on our side.  We do plan on implementing your 
```
add_filter('show_admin_bar', array($cloudflareHooks, 'hideAdminBar'));
```
change as well though.

I really appreciate you taking the time to contribute and I apologize for saying I would merge it.  If you would like I'd love to send you some Cloudflare swag - if you open up a support ticket and reference this PR's URL we can get your mailing address and send you a bunch of stuff.

Thanks,
John

piscis:
Hi John, thanks for letting me know.

Issue #104, opened 2016-10-09
Class 'CloudFlare\IpRewrite' not found


tillkruss:
Plugins version: 3.0.6
Hosting: Heroku

The plugin works fine, but when I execute `wp user list` (wp-cli.org), I get the following error:

![screen shot 2016-10-09 at 3 11 03 pm](https://cloud.githubusercontent.com/assets/665029/19224068/9f7d1a34-8e32-11e6-9a01-3c2c375f50dc.png)

jwineman:
Hey,

Thanks for reporting this!  Does switching 
```
require_once 'vendor/autoload.php';
```
to
```
require __DIR__ . '/vendor/autoload.php';
```
in cloudflare.loader.php fix your issue?

Thanks,
John

tillkruss:
Yes, it does.

jwineman:
Awesome!  Thanks again for reporting this, we'll make sure it gets in the next release.

PR #105, opened 2016-10-11
#104: Composer Autoloader now requires with absolute path to fix bug …

jwineman:
…when plugin was called through WP-CLI.


jwineman:
@thellimist please review/merge.

PR #112, opened 2016-10-19
Purge Post URL, Home, Blog and Terms page on update

brandomeniconi:
Allows WP to purge this URLS on post (also page and CPT) update:

- Post URL
- Homepage
- Page for posts (if set)
- Related tags URLs
- Related categories/taxonomies URLs
- HTTP & HTTPS (only if admin from HTTPS)

Doesn't purge on post autosave.

I've tested this on Google AppEngine.

### DIFF
--- a/cloudflare.loader.php
+++ b/cloudflare.loader.php
@@ -46,4 +46,5 @@ if (is_admin()) {
     // Load Automatic Cache Purge
     add_action('switch_theme', array($cloudflareHooks, 'purgeCache'));
     add_action('customize_save_after', array($cloudflareHooks, 'purgeCache'));
+    add_action('save_post', array($cloudflareHooks, 'purgePage'));
 }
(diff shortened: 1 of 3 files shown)


cloudflare.loader.php
```diff
@@ -46,4 +46,5 @@
     // Load Automatic Cache Purge
     add_action('switch_theme', array($cloudflareHooks, 'purgeCache'));
     add_action('customize_save_after', array($cloudflareHooks, 'purgeCache'));
+    add_action('save_post', array($cloudflareHooks, 'purgePage'));
```
jwineman: assuming `is_admin()` returns true any time they're saving a post correct?

brandomeniconi:
Yes, correct. I've this code live on a news website and it's working as
expected.

2016-10-19 17:40 GMT+02:00 John Wineman <notifications@github.com>:

jwineman:
This is awesome, thank you!

PR #114, opened 2016-10-20
Map active zone to Primary Domain

brandomeniconi:
When Wordpress MU Domain Mapping is enabled, use the primary domain for current blog, instead of default subdomain.

Example scenario:

| wp-admin | PRIMARY DOMAIN | blog_id |
| --- | --- | --- |
| example.com | example.com | 1 |
| it.example.com | www.example.it | 2 |
| fr.example.com | www.example.fr | 3 |

actually this plugin uses the example.com zone for all the domains.

With this PR the active zone is set to the Primary Domain setup in [Wordpress MU Domain Mapping](https://it.wordpress.org/plugins/wordpress-mu-domain-mapping/).

### DIFF
--- a/src/WordPress/WordPressWrapper.php
+++ b/src/WordPress/WordPressWrapper.php
@@ -21,6 +21,10 @@ class WordPressWrapper
 
     public function getSiteURL()
     {
-        return get_site_url();
+        if(function_exists('domain_mapping_siteurl')){
+            return domain_mapping_siteurl('');
+        } else {
+           return get_site_url();
+        }    
     }
 }
--- a/src/WordPress/WordPressWrapper.php
+++ b/src/WordPress/WordPressWrapper.php
@@ -21,7 +21,7 @@ class WordPressWrapper
 
     public function getSiteURL()
     {			
-		$site_url = get_site_url();
+	$site_url = get_site_url();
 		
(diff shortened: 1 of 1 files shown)


src/WordPress/WordPressWrapper.php
```diff
@@ -21,6 +21,10 @@ public function deleteOption($key)
 
     public function getSiteURL()
     {
-        return get_site_url();
+        if(function_exists('domain_mapping_siteurl')){
+            return domain_mapping_siteurl('');
```
thellimist: where does this function exist?

brandomeniconi:
Updated description

jwineman:
Can probably accept this as well even though we aren't officially supporting WordPress Mu.  Need to talk it over with our PM.

brandomeniconi:
Ok, I know Domain Mapping plugin is old and messy, but it's the "official one" from Automattic. This is a quick & dirty solution. Probably a better implementation is to add the domains to the CF Active Zone select in backend and let the user choose the right one.

It's up to you to support Domain Mapping plugin, if not, I found another solution to this issue: disable "Remote login" and "Administration Redirect" from the DM plugin settings. This opens the wp-admin pages on primary domain and not the multisite subdomain.

thellimist:
Thanks!

PR #120, opened 2016-11-04
DES-965 updated cloudflare-frontend 2.6.0

thellimist:


thellimist:
@jwineman Please don't merge. I need to test something

PR #135, opened 2016-11-22
Improve image compression.

XhmikosR:
Everything is still losslessly compressed.

Before: 6,03 KB (6.177 bytes)
After:  5,39 KB (5.520 bytes)


thellimist:
Thanks

PR #139, opened 2016-11-24
Use a space around `.`.

XhmikosR:
This is subjective, but IMO makes it easier to spot where the concatenation is happening.

I also fixed  a small type while at it.

Let me know your thoughts :)

### DIFF
--- a/cloudflare.loader.php
+++ b/cloudflare.loader.php
@@ -1,6 +1,6 @@
 <?php
 
-require_once __DIR__.'/vendor/autoload.php';
+require_once __DIR__ . '/vendor/autoload.php';
 
 use CloudFlare\IpRewrite;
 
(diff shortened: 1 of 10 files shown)


thellimist:
We're using [PSR2](http://www.php-fig.org/psr/psr-2/) coding style guide. Running `composer format` should tell whether it's supported or not.

XhmikosR:
I don't have composer handy atm, so if you could check it out and let me know that would be sweet.

thellimist:
Try to run `composer test`, `composer format` before you push :wink:

XhmikosR:
The thing is that I'm a Windows user and things don't work out of the box with your setup :/

Anyway, I managed to make things work, and this passes `composer format`.

PR #141, opened 2016-11-24
index.php: consistent formatting for inline JS.

XhmikosR:
Note that I kept the bracket notation for properties to match the PHP code.

Only things I'd personally improve, is to explicitly set global variables to the global scope so that it's clear what is what. But that's for another time.


XhmikosR:
Note that I kept the bracket notation for properties to match the PHP code.

Only things I'd personally improve, is to explicitly set global variables to the global scope so that it's clear what is what. But that's for another time.

PR #146, opened 2016-11-24
composer.json: Windows compatibility.

XhmikosR:
Only downside I see from this is when one has `php5` and not `php`, but I think it's symlinked so it should be OK.

Let me know your thoughts.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,8 +22,8 @@
         "simplyadmire/composer-plugins" : "@dev"
     },
     "scripts": {
-        "format": "vendor/bin/phpcs -n --standard=phpcs.xml",
-        "test": "vendor/bin/phpunit",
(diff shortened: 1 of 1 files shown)


XhmikosR:
I added an alternative patch. It requires `composer update` though to have the vendor/bin/*.bat files. If we go with this approach, it would be better to have the `composer update` before this.

Please let me know your thoughts and with which solution you want to go. Either way it works for both Unix and Windows.

XhmikosR:
I went with my initial approach. There are still failures on Windows due to `php-compatibility-install` task.

I wish there was a way to have cross platform scripts in composer.

If anyone has a better idea let me know. Until then, I can at least run `composer test` and `composer format` successfully.

thellimist:
Works on UNIX properly so it seems good

PR #144, opened 2016-11-24
Enforce LF for most filetypes in the repo.

XhmikosR:
This fixes an issue with `composer format` failing on Windows when git's `autocrlf` is set to true and thus the repo line endings are `\r\n`.

PR #149, opened 2016-11-25
Tweak trim_code_for_release.sh.

XhmikosR:
* switch to the dist folder before doing anything; reduces duplication
* remove `-r` when removing compiled.js.map
* remove .gitignore
* remove all test folders from vendor subfolders

Saves ~100KB from the distribution archive.

Looking at v3.0.0 which weighted **4.0 MB**, v3.1.1 weighted **2.5 MB**, and the next release should be **~1.54 MB** if the rest of my PRs are merged. Not bad :)

### DIFF
--- a/scripts/deployment/trim_code_for_release.sh
+++ b/scripts/deployment/trim_code_for_release.sh
@@ -4,27 +4,44 @@
 REPOSITORY_FOLDER="$1"
 
 if [ -z "$REPOSITORY_FOLDER" ]; then
-	echo "Please provide the folder name as argument."
-	exit 1
+    echo "Please provide the folder name as argument."
+    exit 1
 fi
 
-rm -f $REPOSITORY_FOLDER/phpcs.xml
-rm -f $REPOSITORY_FOLDER/phpunit.xml
-rm -f $REPOSITORY_FOLDER/.editorconfig
-rm -rf $REPOSITORY_FOLDER/.git
-rm -rf $REPOSITORY_FOLDER/src/Test/
-rm -rf $REPOSITORY_FOLDER/git-hooks
-rm -rf $REPOSITORY_FOLDER/compiled.js.map
-rm -rf $REPOSITORY_FOLDER/scripts
-rm -rf $REPOSITORY_FOLDER/vendor/bin/
-rm -rf $REPOSITORY_FOLDER/vendor/squizlabs
-rm -rf $REPOSITORY_FOLDER/vendor/phpunit
-rm -rf $REPOSITORY_FOLDER/vendor/php-mock
-rm -rf $REPOSITORY_FOLDER/vendor/johnkary 
-rm -rf $REPOSITORY_FOLDER/vendor/guzzle/guzzle/tests
-rm -rf $REPOSITORY_FOLDER/vendor/guzzle/guzzle/docs
-rm -rf $REPOSITORY_FOLDER/vendor/phpdocumentor
-rm -rf $REPOSITORY_FOLDER/vendor/webmozart
-rm -rf $REPOSITORY_FOLDER/vendor/sebastian/global-state
-rm -rf $REPOSITORY_FOLDER/vendor/wimg
-rm -rf $REPOSITORY_FOLDER/vendor/simplyadmire
+pushd "$REPOSITORY_FOLDER"
+
+rm -f .editorconfig
+rm -f .gitignore
+rm -f .travis.yml
+rm -f compiled.js.map
+rm -f phpcs.xml
+rm -f phpunit.xml
+rm -rf .git/
+rm -rf git-hooks/
+rm -rf scripts/
+rm -rf src/Test/
+rm -rf vendor/bin/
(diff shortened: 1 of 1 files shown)


XhmikosR:
Looking at v3.0.0 which weighted 4.0 MB, v3.1.1 weighted 2.5 MB, and the next release should be ~1.76 MB if my PRs are merged. Not bad :)

XhmikosR:
One thing, I'm wondering: Is composer.json and composer.lock needed for distribution?

scripts/deployment/trim_code_for_release.sh
```diff
@@ -4,24 +4,38 @@
...
-rm -rf $REPOSITORY_FOLDER/vendor/squizlabs
-rm -rf $REPOSITORY_FOLDER/vendor/phpunit
-rm -rf $REPOSITORY_FOLDER/vendor/php-mock
-rm -rf $REPOSITORY_FOLDER/vendor/johnkary 
-rm -rf $REPOSITORY_FOLDER/vendor/guzzle/guzzle/tests
-rm -rf $REPOSITORY_FOLDER/vendor/guzzle/guzzle/docs
-rm -rf $REPOSITORY_FOLDER/vendor/phpdocumentor
-rm -rf $REPOSITORY_FOLDER/vendor/webmozart
-rm -rf $REPOSITORY_FOLDER/vendor/sebastian/global-state
+pushd "$REPOSITORY_FOLDER"
```
thellimist: I don't like `pushd` command. It's not failure proof.

`-rm -f $REPOSITORY_FOLDER/.editorconfig` is a cleaner way.

thellimist:
> One thing, I'm wondering: Is composer.json and composer.lock needed for distribution?

Some users who understand how stuff work can build their own plugin with `composer.json` in their servers. It also gives them a sense of how stuff work. `composer.lock` is to guarantee these users to download the correct content.

It should stay.

scripts/deployment/trim_code_for_release.sh
XhmikosR: I don't see how duplicating `$REPOSITORY_FOLDER` 20 times is cleaner.

XhmikosR:
Right, it should. But will it work with the stuff this script removes?

XhmikosR:
Anyway, I added one more patch. If everyone agrees, let me know in order to squash the patches.

thellimist:
Looks legit.

XhmikosR:
Added a couple more files. Please test a release from this branch just to make sure before merging.

thellimist:
Even though `composer install --no-dev` is a good idea I've tested and it doesn't seem to work as expected. 

```
Script rm -rf vendor/squizlabs/php_codesniffer/CodeSniffer/Standards/PHPCompatibility; cp -rp vendor/wimg/php-compatibility vendor/squizlabs/php_codesniffer/CodeSniffer/Standards/PHPCompatibility handling the php-compatibility-install event returned with an error


  [RuntimeException]
  Error Output: cp: vendor/wimg/php-compatibility: No such file or directory
```

I've refreshing `vendor/` file a `composer.lock` but couldn't find a error. Could you return to the old where we delete every file one by one. Also `guzzle/guzzle` should be removed

https://github.com/cloudflare/Cloudflare-WordPress/pull/157

XhmikosR:
`composer --no-dev` works right in this case. I don't see why you'd need `vendor/wimg/php-compatibility` in the distribution package.

So, IMO the composer scripts should change.

XhmikosR:
Anyway, check the last patch I pushed.

XhmikosR:
I'm thinking, this should be merged if it works, and then someone should look into #144; having a .gitattributes file we can use `export-ignore` for the files/folders you don't need in a release, and use [git archive](https://git-scm.com/docs/git-archive) to create the zip.

XhmikosR:
I'm thinking, this should be merged if it works, and then someone should look into #144; having a .gitattributes file we can use `export-ignore` for the files/folders you don't need in a release, and use [git archive](https://git-scm.com/docs/git-archive) to create the zip.

thellimist:
The code seems good. Can you also add `rm -rf ./vendor/guzzle` then I'll merge

XhmikosR:
Isn't guzzle required by cloudflare/cloudflare-plugin-backend?

That is why I definitely prefere the `composer --no-dev` approach; it's much less error-prone...

jwineman:
We switched to `wp_remote_request()` in https://github.com/cloudflare/Cloudflare-WordPress/pull/157

XhmikosR:
But it's referenced as a dependency from cloudflare/cloudflare-plugin-backend in composer.json.

jwineman:
Guzzle is being included becuase its a dependency of `cloudflare/cloudflare-plugin-backend`.  Eventually we'll move the dependency out of that repository and only include it in the plugins that use it (Magento/cPanel).

XhmikosR:
All right, if you guys insist, you should know better :)

PR #134, opened 2016-11-22
Make API requests through WP's native WP_HTTP class. Drop Guzzle/cURL dependency

brandomeniconi:
As discussed in #113 this PR allows the plugin to use WordPress native HTTP class.

PR #153, opened 2016-11-25
Add .travis.yml for Travis CI testing.

### LINKED ISSUE #143: Set up a CI system like Travis CI
Now, there is no automated way as far as I can tell, so adding Travis CI will make sure things work all the time.

XhmikosR:
Some notes:

1. php/mock requires PHP >= 5.5. I'm not sure what to do in order to test all the supported by cloudflare-WordPress plugin PHP versions, so I skipped 5.3 and 5.4. IMO we should definitely test all supported versions to prevent accidental breakage.
2. I left `fast_finish` out and HHVM.

Finally. before merging this, you will need to add Travis CI to the repo and enable "Build only if .travis.yml is present" in the Travis project's settings.

Until then you can have a quick look https://travis-ci.org/XhmikosR/Cloudflare-WordPress/builds

Fixes #143.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,18 @@
+language: php
+
+php:
+    #- '5.3'
+    #- '5.4'
+    - '5.5'
+    - '5.6'
+    - '7.0'
+
+#matrix:
+#    fast_finish: true
+
+install:
+    - composer install
+
+script:
+    - composer format
+    - composer test
--- a/README.md
+++ b/README.md
@@ -1,5 +1,7 @@
 # Cloudflare’s Plugin for WordPress
 
+[![Build Status](https://travis-ci.org/cloudflare/Cloudflare-WordPress.svg?branch=master)](https://travis-ci.org/cloudflare/Cloudflare-WordPress)
+
 Cloudflare’s WordPress plugin brings all of the benefits of Cloudflare into your WordPress dashboard for configuration, including a one-click application of default settings specifically optimized for WordPress.
 
 By enabling Cloudflare on your WordPress website, you’ll find performance and security gains such as doubling your page load speeds, DDoS protection, web application firewall with WordPress-specific rulesets, free SSL, and SEO improvements.


jwineman:
Thanks!

XhmikosR:
@jwineman: how about PHP < 5.5?

jwineman:
I don't know if there is anything we can do from a unti testing perspective since PHPUnit requires 5.5.  We could run PHPCS in compatibility mode for 5.3 and 5.4.  Something like:

```
phpcs --standard=PHP53Compatibility 
```

I think most of the 5.3/5.4 issues are in Guzzle so might be worth holding off on that until we merge https://github.com/cloudflare/Cloudflare-WordPress/pull/134

XhmikosR:
But how will `composer install` work when some dependencies list PHP >= 5.5 as a requirement?

Also, might be better to have a ticket to track this issue so it doesn't get lost.

jwineman:
`composer format` runs PHPCS to lint.  We can also have it run PHPCS again to check compatibility against 5.3 and 5.4 using that command I posted earlier.  This way the linter will fail if you use PHP functions not supported in those versions. Then we continue to test on 5.5+ and assume that if the unit tests pass and we're not using any functions not in 5.3 and 5.4 that it will work on those versions.  Obviously its not 100% certainty but I think its as close as we can get.

XhmikosR:
But the failure happens in `composer install` step, not any further.

jwineman:
PHPCS will run in 5.5+ but check for code compatibility on 5.3/5.4, this shouldn't cause `composer install` to fail right?

I don't think we can actually test on these lower environments since our tooling requires higher versions of PHP.  This is as close as I can get off the top of my head.

XhmikosR:
The build will fail when running `composer install` due to the requirement some deps have on PHP >= 5.5. I'm not sure how we can work around this issue without forking those projects.

XhmikosR:
Also, should the `post-install-cmd` be in composer.json anymore?

jwineman:
Yes, I understand that.  We don't want to fork dependencies so what I'm suggesting is using

https://github.com/wimg/PHPCompatibility

To ensure that there aren't any PHP functions in our code that aren't compatible with 5.3, 5.4. 

No the post-install-cmd and the git-hooks can be removed. I have that as a TODO.

XhmikosR:
As long as you manage to install the deps with `composer install` that seems like a good solution.

Just keep it somewhere so that you don't forget about it.

jwineman:
Yeah I think we can make it part of the `composer format` command, I just need to play with it.  Just wanted to accept the PR initially so we're at least running the unit tests for now.

jwineman:
Tracking internally PI-919

PR #156, opened 2016-11-29
Thellimist/pi 789 remove page rules from Apply Default Settings

thellimist:


thellimist:
Related:
https://github.com/cloudflare/cloudflare-plugin-backend/pull/30

PR #157, opened 2016-11-30
WP Client API uses WP_HTTP class instead of Guzzle.

### LINKED ISSUE #113: Use WP_HTTP instead of Guzzle to make API requests
Hi,
we have a lot of WP websites behind CF and running on Google AppEngine. GAE uses his own library (URLFetch) to make external HTTP calls and hasn't cURL available by default.

We have also few corporate WP installation that doesn't have direct access to internet and they must send requests through a proxy.

The AppEngine Plugin and the `define('WP_PROXY_HOST', '0.0.0.0');` config rules allows WP to manage these situations, but because this plugin use Guzzle instead of the WP_HTTP class, it will not work out of the box in this environments and requires manual patches.

Could this plugin be updated to use the WP_HTTP Class to better follow the WP guidelines?

Thanks for the great work.

XhmikosR:
Removes the php-curl dependency.

Fixes #113.

This is #134, formatted, rebased and squashed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,10 +65,6 @@ Yes, on install and activation the plugin, first time users will be asked to ent
 
 Yes, Cloudflare works with, and helps speed up your site even more, if you have Varnish enabled.
 
-### Why am I getting “Cloudflare plugin requires php5-curl to be installed” error?
-
-Make sure that the php5-curl extension is installed on your system.
-
 ## Support
 
 ### Visit Our Knowledge Base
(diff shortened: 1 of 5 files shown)


XhmikosR:
@thellimist @jwineman: can you please have a look at my PRs? It's annoying having to rebase them :/

thellimist:
@XhmikosR Thanks for fixing the the conflictions.

jb510:
So... the Guzzle dependency was removed, and composer.json updated...  but Guzzle is still checked in.  Is there a reason it hasn't been removed?

jwineman:
Good catch @jb510 - addressing this in https://github.com/cloudflare/Cloudflare-WordPress/pull/174.

PR #161, opened 2017-01-18
Purge both https and http URLs on content modification

brandomeniconi:
After a post update, actually the plugin purges only the URLs with the same scheme of the author admin panel. This appends the URLs with https if the admin is in http and vice-versa.


thellimist:
Is it possible to have admin panel not using SSL but the website is using SSL (or vice-versa)? I don't know a case where this PR would be useful. 

There might be an overhead of purging additional ~15+ links which are not valid. Can we detect if the website is using SSL and just purge the correct links?

brandomeniconi:
My most common situation is having admin under SSL, but the website accessible both via https and http. One reason to do this is because WordPress saves absolute links to images in the post content and if the site was managed via HTTP, there will be a lot of posts with http:// src and srcset images in it. So if a user is now redirected (from old SERP result, Facebook, Referrer, etc link) to HTTPS version of page, he will not see images due to browser CSP. I know that Cloudflare has an option to rewrite content to HTTPS but last time I used it it doesn't support the `srcset` image attribute. Also some old (bad) WIFI Captive Portals struggle to manage their redirect via HTTPS. So in my experience actually it's good to have both schemes enabled, obviously with HTTPS version as canonical URL 😄 

Also WP Codex has a page that explains it: https://codex.wordpress.org/Administration_Over_SSL

The vice versa situation is lot less common, I agree, and it's also bad. Maybe we can try to guess theese situations to avoid unnecessary overhead:

```php
if(force_ssl_admin()){
  $listofurls = array_merge( $listofurls, str_replace( 'https://', 'http://', $listofurls ) );
} elseif(!is_ssl() && force_ssl_content()) {
  $listofurls = array_merge( $listofurls, str_replace( 'http://', 'https://', $listofurls ) );
}

thellimist:
> the website accessible both via https and http. 
Correct me if I'm wrong; `is_ssl()` will return `true` in this case. If so the the logic seems good enough. 

When you add the logic to avoid unnecessary overhead and fix the linter I'll accept the PR.

Thanks

brandomeniconi:
Updated.
> Correct me if I'm wrong; is_ssl() will return true in this case. If so the logic seems good enough.

If you're referring to `!is_ssl()` in the second condition, it will return `true` if the user is updating post from a non SSL admin.

thellimist:
Thanks

PR #163, opened 2017-01-31
Purge all post related taxonomies terms

brandomeniconi:
Actually the plugin only purges category and tag terms, but there are many situations where a post has public custom taxonomies that needs to be purged on post update.

This will purge the terms pages for all taxonomies (including tags and categories) registered with the post's post_type.


thellimist:
Thanks

PR #166, opened 2017-02-02
Thellimist/pi 993

thellimist:


thellimist:
@jwineman Merged for beta-testing

src/WordPress/HTTP2ServerPush.php
```diff
@@ -1,9 +1,9 @@
 <?php
 
-// TODO: Get rid of $GLOBALS and use static variables
-namespace CF\Hooks;
-
-use CF\WordPress\Utils;
+// TODO: 
+// Get rid of $GLOBALS and use static variables
+// Make class functions non-static
+namespace CF\WordPress;
```
jwineman: @thellimist Is this resolved?

thellimist: It is used here: https://github.com/cloudflare/Cloudflare-WordPress/blob/c30fc3e478c9662b46e6fe883afae3d2d8b6ba38/src/WordPress/HTTP2ServerPush.php#L57

Also everything is static which is not how we want to write code. They are still TODO's.

thellimist: @jwineman

PR #172, opened 2017-02-10
Thellimist/pi 1002

thellimist:


jwineman:
Build failed.

src/WordPress/Hooks.php
```diff
@@ -18,8 +18,8 @@ class Hooks
 
     public function __construct()
     {
-        $this->config = new Integration\DefaultConfig('[]');
-        $this->logger = new Integration\DefaultLogger(false);
+        $this->config = new Integration\DefaultConfig(file_get_contents(CLOUDFLARE_PLUGIN_DIR.'config.js', true));
+        $this->logger = new Integration\DefaultLogger($this->config->getValue('debug'));
```
jwineman: Does this actually turn on debug logging?  I feel like when we switched to the PSR logger we removed this because you had to turn it on some other way.

thellimist: I tested it with the change. It does work

src/Test/WordPress/HooksTest.php
```diff
@@ -42,9 +42,10 @@ public function setup()
         $this->mockDefaultIntegration = new DefaultIntegration($this->mockConfig, $this->mockWordPressAPI, $this->mockDataStore, $this->mockLogger);
 
         $this->hooks = $this->getMockBuilder('\CF\WordPress\Hooks')
-            ->setMethods(array('setHooks'))
+            ->disableOriginalConstructor()
+            ->setMethods(array('__construct')) // This is a hack to make the tests work
```
jwineman: `->disableOriginalConstructor()` disables the original constructor and then you mock one with `->setMethods(array('__construct'))`.  Do you need the constructor?  If so PHPUnit allows you to mock with it.

thellimist: I needed to avoid constructor being called but not mocking other functions. Mocking a random function in this case `'__construct'` solved it.

jwineman: Oh wait this is the HookTest, why are we mock it at all?  If this is the class under test it should be real.

thellimist: I didn't get the question. We are mocking the objects and checking if the class Hook calls the correct objects with the correct functions.

jwineman: When you write tests for a class you don't mock the class that is being tested.  Since this is HooksTest.php we shouldn't be mocking Hooks, we should be instantiating the real one.

thellimist: > we should be instantiating the real one.
The error was caused by instantiating. This was a really quick fix which avoided calling constructor but doesn't mock anything else. 

Mocking these [two](https://github.com/cloudflare/Cloudflare-WordPress/blob/master/src/WordPress/Hooks.php#L21) didn't seem straight forward instead I mocked the constructor to avoid it being called.

jwineman: This is why you pass your dependencies in the constructor instead of `new`ing them _in_ the constructor.  We need to modify the constructor to accept args for all those values, then pass in mocked versions of those values instead of the real thing.

thellimist: I agree. Followed PI-1009

PR #173, opened 2017-02-20
Check if constant CLOUDFLARE_HTTP2_SERVER_PUSH_ACTIVE is defined

brandomeniconi:
If the named constant `CLOUDFLARE_HTTP2_SERVER_PUSH_ACTIVE` is not defined in` wp-config.php` PHP uses it's name as value and enables the feature by default. It also shows warnings. 

Better check if it's defined before.


thellimist:
Good catch thanks!

PR #174, opened 2017-02-23
Updating vendor/ now that composer.json no longer depends on Guzzle.

jwineman:
@thellimist do I need to update any of the build scripts since they might be removing stuff that is no longer in the vendor folder?

`composer.lock` also updated as part of this PR.


thellimist:
Release process doesn't depend on vendor folder for we can ignore those scripts.

In composer.json

```
    "scripts": {
        "format": "php vendor/squizlabs/php_codesniffer/scripts/phpcs -n --standard=phpcs.xml", 
        "test": "php vendor/phpunit/phpunit/phpunit", 
        "post-install-cmd": "@php-compatibility-install", 
        "post-update-cmd": "@php-compatibility-install", 
        "php-compatibility-install": "rm -rf vendor/squizlabs/php_codesniffer/CodeSniffer/Standards/PHPCompatibility; cp -rp vendor/wimg/php-compatibility vendor/squizlabs/php_codesniffer/CodeSniffer/Standards/PHPCompatibility"
    }, 
```

I don't remember what we did with `php-compatibility-install`. Others seem fine

jwineman:
Build is failing cause our test dependencies require higher versions of PHP than we test on.  I'm going to modify run the build scripts on the vendor folder and commit it so we don't have to run them before each release.

jwineman:
Depends on https://github.com/cloudflare/cloudflare-plugin-backend/pull/33

thellimist:
Do you need to release backend and update composer file backend version?

jwineman:
Yes, but I think we might need more changes to the backend so I'm going to hold off until I know were done changing it.

XhmikosR:
Maybe it's time to stop supporting the ancient PHP versions? This would simplify things a lot.

jwineman:
We try to match what WordPress supports which is why we support ~5.3+ (whenever they introduced namespace support).  Still want to run the tests on as many versions as possible.  I can get this working on 5.5 again just need to mess with the `composer.json` more.

thellimist:
@jwineman FYI
Issue was phpunit-speedtrap updated from version 1.0.1 to 1.0.2 and changed the min php version in composer.json where as they should have updated to 1.1.0 to avoid breaking dependencies

PR #176, opened 2017-03-10
Thellimist/pi 1035

thellimist:


src/WordPress/WordPressClientAPI.php
```diff
@@ -93,91 +93,99 @@ public function createPageRule($zoneId, $body)
...
 
-        $mergedResponse = null;
-
-        $currentPage = 1;
-        $totalPages = 1;
-
-        while ($totalPages >= $currentPage) {
-            // Enable pagination
-            if ($isPaginatable) {
-                $params['page'] = $currentPage;
```
thellimist: The bug was caused here and at line 97: `$url = add_query_arg($request->getParameters(), $this->getEndpoint().$request->getUrl());` function.

While refactoring to guzzle to missed `$params` variable which doesn't exist.

Other thing we missed was in Guzzle the args were suppose to be sent in `send` function where as in wordpress it's suppose to be sent in `add_query_arg`. We weren't setting the `page` parameter of the request so it wasn't grabbing pages after 1

PR #177, opened 2017-03-10
PI-1022 fix activezoneselector css

thellimist:
I'm planing to update the frontend in this PR. Let me know if it looks fine


stylesheets/hacks.css
```diff
@@ -4,6 +4,13 @@
  * everything here.
  */
 
+/* fix active-zone-selector being to far from the logo. The selector
+ * still has bad css when using mobile or small screen view.
+ */
+.active-zone-selector__singlezone {
+    margin-left: auto !important;
```
jwineman: Are you using `!important` because you need to override inline styles?  Does this change apply to cPanel and Magento as well?  Would it make sense to move it inline in `ActiveZoneSelector.js` in `cloudflare-plugin-frontend`?

thellimist: This PR basically undo's this [commit](https://github.com/cloudflare/cloudflare-plugin-frontend/commit/f21842fda4dd205e51b51745b4b74d9d633ca00b) for WordPress only. The styles are messed up in general. Magento and CPanel are better with the `margin-left` where as WordPress looks extremely ugly with it. By using `!important` we allow Magento and CPanel to work but fix WordPress.

> Would it make sense to move it inline in ActiveZoneSelector.js in cloudflare-plugin-frontend?

No since it's a plugin specific fix.

jwineman: Okay.  As a general rule `!important` is a mortal sin but this is `hacks.css` so ¯\_(ツ)_/¯

PR #179, opened 2017-04-13
PI-1062 add purge cache button to wp admin bar

thellimist:
There are multiple things going on here. We don't have access to anything outside of plugin page level. We need

- Client v4 API Key
- Check if "cache everything" page rule exists and show to only those users

============================================================

1) Once we have access to API Key, we make page rule request. 
2) To make a page rule request we need to have `zoneId`. That's why I'm caching the zoneId every time the user enters the plugin page. 
3) Once we know the zoneId, we can make a the request and figure out if there is any page rule related which has "Cache Everything" action in it.  

    - TODO: We need to make sure not any page rule but only this zone should have it.
    - TODO: Currently all these checks will run for every single request to the page. We need to limit that to admin login (not admin page)
    - TODO: We need to make sure not to request the API every time the user hangouts in admin page. We need to cache the result of `isCacheEverythingPageRuleExists` function and once in a while (1 day seems good I think. Any ideas?) we need double check if this page rule still exists. 

4) We know cache everything page rule exists, purge cache and alert the user

### DIFF
--- a/src/WordPress/DataStore.php
+++ b/src/WordPress/DataStore.php
@@ -11,6 +11,7 @@ class DataStore implements DataStoreInterface
     const API_KEY = 'cloudflare_api_key';
     const EMAIL = 'cloudflare_api_email';
     const CACHED_DOMAIN_NAME = 'cloudflare_cached_domain_name';
+    const CACHED_ZONE_ID = 'cloudflare_cached_zone_id';
 
     protected $wordPressWrapper;
 
@@ -59,6 +60,18 @@ class DataStore implements DataStoreInterface
     }
 
     /**
+     * @return bool
+     */
+    public function getClientAPICredentialsExist()
+    {
+        if ($this->getClientV4APIKey() && $this->getCloudFlareEmail()) {
+            return true;
+        }
+
+        return false;
+    }
+
+    /**
      * @return client v4 api key for current user
      */
     public function getClientV4APIKey()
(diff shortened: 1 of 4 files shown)


thellimist:
> TODO: We need to make sure not any page rule but only this zone should have it.

This TODO is kind of tricky. We need to do string manipulation 😞

src/WordPress/DataStore.php
```diff
@@ -59,6 +60,18 @@ public function getHostAPIUserUniqueId()
     }
 
     /**
+     * @return bool
+     */
+    public function getClientAPICredentialsExist()
```
jwineman: I might rename this to `clientAPICredentialsExist()`. "get" implies it returns something.

src/WordPress/Hooks.php
```diff
@@ -242,4 +246,63 @@ public function http2ServerPushInit()
...
+            'id' => 'cloudflare-purge-cache',
+            'title' => 'Cloudflare Purge Cache',
+            'href' => '#'
+        );
+
+        $wpAdminBar->add_node($args);
+    }
+
+    public function shouldShowAdminBarMenuButton() {
+        // TODO add showadminbar cache to not query everytime the page changes
```
jwineman: So are you trying to merge this with these TODOs or is this no longer a TODO?

thellimist: All of the TODO's are still TODO's. I wanted to push and show the way I'm taking before going further in depth. (read the PR description)

jwineman: Can we test this?

src/WordPress/WordPressClientAPI.php
```diff
@@ -86,6 +87,42 @@ public function createPageRule($zoneId, $body)
...
+        return $response;
+    }
+
+    /**
+     * @param $zoneId
+     *
+     * @return bool
+     */
+    public function isCacheEverythingPageRuleExists($zoneId) {
+        $activePageRuleList = $this->getActivePageRuleList($zoneId);
```
jwineman: Can we test this?

src/WordPress/Hooks.php
thellimist: I assume you meant writing test. Will do

thellimist:
I'm closing this PR since we think it doesn't benefit majority of our user base.

PR #182, opened 2017-05-05
PI-812: We read request JSON for cloudflare plugin requests after the…

jwineman:
`php://input` can only be read once < PHP 5.6 and PUT requests can't be rewound.  This means we need to try to read the `php://input` stream as early in the script execution as possible.  Wordpress outlines the order in which hooks execute [here](https://codex.wordpress.org/Plugin_API/Action_Reference).  The order is:

- `muplugins_loaded`
- `registered_taxonomy`
- `registered_post_type`
- `plugins_loaded`

`muplugins_loaded` is only available in WordPress 4.x, `register_taxonomy` doesn't list the version it was introduced in, `registered_post_type` doesn't have docs which leaves us with `plugins_loaded`.

This bug was surfaced because a plugin was reading Cloudflare requests after hooking into `setup_theme` which executes after `plugins_loaded`.

### DIFF
--- a/cloudflare.loader.php
+++ b/cloudflare.loader.php
@@ -23,10 +23,6 @@ if ($isCf) {
 // Initiliaze Hooks class which contains WordPress hook functions
 $cloudflareHooks = new \CF\WordPress\Hooks();
 
-/*
- * php://input can only be read once before PHP 5.6, try to grab it ONLY if the request
- * is coming from the cloudflare proxy
- */
 add_action('plugins_loaded', array($cloudflareHooks, 'getCloudflareRequestJSON'));
 
 // Enable HTTP2 Server Push
--- a/cloudflare.loader.php
+++ b/cloudflare.loader.php
@@ -24,7 +24,7 @@ if ($isCf) {
 $cloudflareHooks = new \CF\WordPress\Hooks();
 
 /*
- * php://input can only be read once before PHP 5.6, try to grab it IF the request
+ * php://input can only be read once before PHP 5.6, try to grab it ONLY if the request
  * is coming from the cloudflare proxy
  */
 add_action('plugins_loaded', array($cloudflareHooks, 'getCloudflareRequestJSON'));
--- a/cloudflare.loader.php
+++ b/cloudflare.loader.php
@@ -23,6 +23,12 @@ if ($isCf) {
 // Initiliaze Hooks class which contains WordPress hook functions
 $cloudflareHooks = new \CF\WordPress\Hooks();
 
+/*
+ * php://input can only be read once before PHP 5.6, try to grab it IF the request
+ * is coming from the cloudflare proxy
+ */
+add_action('plugins_loaded', array($cloudflareHooks, 'getCloudflareRequestJSON'));
+
 // Enable HTTP2 Server Push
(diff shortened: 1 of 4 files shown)


src/WordPress/Hooks.php
```diff
@@ -250,4 +252,11 @@ public function http2ServerPushInit()
     {
         HTTP2ServerPush::init();
     }
+
+    public function getCloudflareRequestJSON()
+    {
+        if ($_GET['action'] === 'cloudflare_proxy') {
```
thellimist: When a request is PATCH or DELETE does `$_GET['action']` still be occupied with `cloudflare_proxy`?

cloudflare.loader.php
```diff
@@ -23,6 +23,12 @@
 // Initiliaze Hooks class which contains WordPress hook functions
 $cloudflareHooks = new \CF\WordPress\Hooks();
 
+/*
+ * php://input can only be read once before PHP 5.6, try to grab it IF the request
+ * is coming from the cloudflare proxy
+ */
+add_action('plugins_loaded', array($cloudflareHooks, 'getCloudflareRequestJSON'));
```
thellimist: Let's be more clear by changing `try to grab it IF the request` to `try to grab it ONLY if the request`.

src/WordPress/Hooks.php
thellimist: Let's put `cloudflare_proxy` into a global variable in Hooks.php. It is being used in [index.php](https://github.com/cloudflare/Cloudflare-WordPress/blob/master/index.php#L49) as well. 

`const CLOUDFLARE_PROXY = "cloudflare_proxy"`

src/WordPress/Proxy.php
```diff
@@ -102,7 +102,7 @@ public function createRequest()
      */
     public function getJSONBody()
     {
-        return file_get_contents('php://input');
+        return $GLOBALS[Hooks::CLOUDFLARE_JSON];
```
thellimist: This is not a straightforward thing to do. `getJSONBody()` and `getCloudflareRequestJSON()` needs docs explaining what this is. It's so hard for a new dev to grasp why we do something like this and it should be clear in the code with where it's being used and why.

src/WordPress/Hooks.php
```diff
@@ -16,6 +16,8 @@ class Hooks
     protected $logger;
     protected $proxy;
 
+    const CLOUDFLARE_JSON = "CLOUDFLARE_JSON";
```
thellimist: instead of double quote, let's use single quote to match our style. I'm not sure why the linter didn't get this..

src/WordPress/Proxy.php
jwineman: You need `getJSONBody()` for [tests](https://github.com/cloudflare/Cloudflare-WordPress/blob/56dda99cdd915521a9b1982cd7c94dbaad4df069/src/Test/WordPress/ProxyTest.php#L44).  How would you refactor?

src/WordPress/Hooks.php
jwineman: All HTTP verbs hit the same URL with the `action` set because thats how WordPress determines which AJAX handler to use.

src/WordPress/Proxy.php
thellimist: Ohh, I meant just comment on those two functions what why they do this

PR #183, opened 2017-05-27
COM-32 :: Add logging over oversized Server Push HTTP headers

IcyApril:


thellimist:
Normally we use `LoggerInterface` to logs. It'd be a hassle  to use it so `error_log` is fine imo

jwineman:
Our WordPress logger logs to `error_log` anyway too.

PR #185, opened 2017-06-26
PI-1193 add try catch block for IpRewrite

thellimist:


cloudflare.loader.php
```diff
@@ -10,14 +10,18 @@
...
-        $_SERVER['HTTPS'] = 'on';
+try {
+    $ipRewrite = new IpRewrite();
+
+    $isCf = $ipRewrite->isCloudFlare();
+    if ($isCf) {
+        // Fixes Flexible SSL
+        if (isset($_SERVER['HTTP_X_FORWARDED_PROTO']) && $_SERVER['HTTP_X_FORWARDED_PROTO'] == 'https') {
+            $_SERVER['HTTPS'] = 'on';
+        }
```
thellimist: Only `IpRewrite` function throws exception. I included the whole block under try catch to make it simpler to understand it's a block.

cloudflare.loader.php
```diff
@@ -10,14 +10,18 @@
...
     }
+} catch (\RuntimeException $e) {
+    error_log($e->getMessage());
```
thellimist: If the user has bad configuration, their logs will be full of this message since every request runs this bit of code including non-admin parts.

jwineman: Thats okay.  If they look at their logs with any frequency they'll disable the plugin or fix their environment.

PR #187, opened 2017-06-28
PI-1198: Making consts in Hooks.php more specific to fix bug.

jwineman:
I assume WordPress 4.8 introduced introduced a constant called `WP_AJAX_ACTION` with a different value from `Hooks::WP_AJAX_ACTION`.  

By using `self` instead of the global `const` we ensure we use our value.

### DIFF
--- a/src/WordPress/Hooks.php
+++ b/src/WordPress/Hooks.php
@@ -257,12 +257,12 @@ class Hooks
     /*
      * php://input can only be read once before PHP 5.6, try to grab it ONLY if the request
      * is coming from the cloudflare proxy.  We store it in a global so \CF\WordPress\Proxy
-     * can act on the request body later on in the script execution. 
+     * can act on the request body later on in the script execution.
      */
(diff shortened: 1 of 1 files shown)


thellimist:
https://github.com/cloudflare/Cloudflare-WordPress/blob/master/index.php#L50 Would this create a problem? I'd suggest changing the variable name to avoid the problem.

thellimist:
Does `CLOUDFLARE_JSON` create a problem as well? would [this](https://github.com/cloudflare/Cloudflare-WordPress/blob/2a991b9016341db136352e5ef630521202be745f/src/WordPress/Proxy.php#L108) create a problem

jwineman:
We just can't access constants without a scope anymore.  In this case `Hooks::` is a scope so its fine.

jwineman:
@thellimist just merge this when you wake up and start the release.  I tested and didn't see any other issues.  Please do the same pre-release testing as well though.

PR #188, opened 2017-06-29
PI-1186 add empty userConfig.js

thellimist:
This is not an ideal solution. To avoid getting the following error I'll add the file for now. If you can think of a solution where we can avoid adding this file and avoiding the error let me know

screen shot 2017-06-29 at 11 20 44


jwineman:
You can't commit `userConfig.js`.  Then their changes will get overwritten every time they update the plugin which defeats the point of adding this file in the first place.  The 404 showing up in `console.log` is unavoidable I think.

thellimist:
Let's talk about this via chat

Issue #192, opened 2017-07-03
Provide filter to hook into purgeCacheByRelevantURLs results


piscis:
we would like a way to purge custom urls based on the postID provided to `Wordpress\Hooks->purgeCacheByRevelantURLSs`.

I wonder if you would accept a PR that basically filters the result of `Hooks->getPostRelatedLinks($postId)` and give us a chance to add some more URLs to the list of URLs the plugin should purge.

**Example:**

      return apply_filters('cf_get_post_related_links', $listofurls, $postId);

thellimist:
Sounds good. We try handle uniqueness of the filters with namespaces. Instead of using `cf_get_post_related_links` please use the naming we have used. 

Thanks

piscis:
sorry I looked at the plugins source and did not find the appropriate namespace to use.
For example I found `apply_filters('http2_link_preload_src', $src);`  in `public static function http2LinkPreloadHeader($src)` but it seams to not use any namespace. 

Could you please tell me how I should declare the filter or point me to an example? 

Right now I declared the filter this way:

     return apply_filters('cf_get_post_related_links', $listofurls, $postId);

Thanks

thellimist:
In [cloudflare.loader.php](https://github.com/cloudflare/Cloudflare-WordPress/blob/master/cloudflare.loader.php) you can see custom actions being added with namespaces.

I haven't tested but the code below should work.

```
$cloudflareHooks = new \CF\WordPress\Hooks();
add_action('cloudflare_get_post_related_links', array($cloudflareHooks, 'getPostRelatedLinks'), $listOfUrls, $postId);
```

Having `cloudflare` prefix rather than `cf` would avoid clashing with other Hooks which may use `cf` prefix. 

We can discuss better after the PR. In the description of the please let us know how you're using this Hook.

jwineman:
@piscis,

To expand on this a little more:

The reason WordPress best practices suggest namespacing plugin specific functions is because those functions all live in the global namespace and that is the easiest way to avoid collisions.  Since  the functions that handle cache purging live inside [`CF\WordPress\Hooks`](https://github.com/cloudflare/Cloudflare-WordPress/blob/master/src/WordPress/Hooks.php) the function names don't need to be namespaced because they only exist inside the Hooks class scope and therefore can't be overridden by other plugins.

I'm not familiar with the filter but I think what you would want instead of
```
return apply_filters('cf_get_post_related_links', $listofurls, $postId);
```
[...]

PR #194, opened 2017-07-06
Add a filter to modify the list urls purged by wordpress via CloudFlare API

piscis:
Adds a filter to the return value of getPostRelatedUrls to inject custom pages to the list of URLs that get purged when a post gets modified.

With this filter a user could hook into the list of urls send to the CF Api and modify them before the get send to the CF API.

We currently use this with the fork in production. Our filter that modifies the list of URLs looks like this:

      <?php
         function adjust_cloudflare_purge_urls($listOfUrls, $postId) {

           $listOfUrls[] = 'http://example.com/custom-entpoint-2-not-related-to-post';
           $listOfUrls[] = 'http://example.com/custom-entpoint-2-not-related-to-post';

           return $listOfUrls;
        }

        add_filter('cf_get_post_related_links', 'adjust_cloudflare_purge_urls', 10, 2);

This PR is related to #192

### DIFF
--- a/cloudflare.loader.php
+++ b/cloudflare.loader.php
@@ -59,7 +59,7 @@ $cloudflarePurgeEverythingActions = array(
 );
 
 foreach ($cloudflarePurgeEverythingActions as $action) {
-    add_action($action, array($cloudflareHooks, 'purgeCacheEverything'));
+    add_action($action, array($cloudflareHooks, 'purgeCacheEverything'), PHP_INT_MAX);
 }
 
 $cloudflarePurgeURLActions = array(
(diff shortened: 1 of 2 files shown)


thellimist:
I see how you're trying to do. I thought it was another way in the issue. Sorry for the confusion.

I'd suggest: 
- changing the name from `cf_get_post_related_links` to `cloudflare_purge_by_url`. 
- moving the filter from `getPostRelatedLinks` to [here](https://github.com/cloudflare/Cloudflare-WordPress/blob/master/src/WordPress/Hooks.php#L154). This way we can ensure `getPostRelatedLinks` function doing only one job.
- add a comment [here](https://github.com/cloudflare/Cloudflare-WordPress/blob/master/cloudflare.loader.php#L65) explaining people that they can filter purging.

This is a nice addition. Thanks

piscis:
As mentioned by @jwineman in #195 I'm going to close this one and recreate a pull-request with the suggestions mentioned by @thellimist due to the issues you having with travis-ci

PR #195, opened 2017-07-06
Adjust default purge hooks priority

piscis:
Set the default priority for purge hooks to `PHP_INT_MAX` the gives third party plugins the chance to run before CF cache is purged. This should be backwards compatible till PHP 5.0.5.

The reason for this is that if a third party plugin hooking into the same actions like the CF plugin does it could potentially create a race condition.

**For example:**
- plugins that clear NGINX cache
- plugins that generate a static version of a page

Both example could could take some time to finish and have the potential to deliver style cache results if the run on the same or lower priority.

### DIFF
--- a/cloudflare.loader.php
+++ b/cloudflare.loader.php
@@ -59,7 +59,7 @@ $cloudflarePurgeEverythingActions = array(
 );
 
 foreach ($cloudflarePurgeEverythingActions as $action) {
-    add_action($action, array($cloudflareHooks, 'purgeCacheEverything'));
+    add_action($action, array($cloudflareHooks, 'purgeCacheEverything'), PHP_INT_MAX);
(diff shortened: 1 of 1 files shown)


jwineman:
Thanks for the PR!  This is great.

@thellimist - Any idea why Travis CI isn't running on forked PRs?  I tried to manually trigger a build and couldn't.

jwineman:
We had Travis set to only build on (local) branch updates, not PR updates.  Now I just need to figure out how to trigger builds for these PRs.

jwineman:
@piscis Sorry can you decline and repush this PR so CI runs?  Its not letting me rerun from the dashboard.

piscis:
ok closing and recreate a pull-request

PR #197, opened 2017-07-15
Adding  a new filter cloudflare_purge_by_url

piscis:
To modify the list of URLs that eventually get purged by Cloudflare through the wordpress plugin a new filter was introduced. The is related to the earlier PR #194 and has the suggesting made by @thellimist


thellimist:
This is great. Thanks!

PR #206, opened 2017-12-12
purge URLs on page/post update

### LINKED ISSUE #200: Automatic cache purge doesn't work when switching a page from public => private
This plugin is usually awesome and works well for purging the CF cache when a post is updated. However it doesn't do it if you change the visibility of an existing page from public to private-- the cached public version stays around.

(I imagine this bug is because it thinks "oh, I don't need to purge the cache for this private post" without taking into consideration that it used to be public a moment ago)

macbookandrew:
Fixes #200

### DIFF
--- a/cloudflare.loader.php
+++ b/cloudflare.loader.php
@@ -83,6 +83,7 @@ $cloudflarePurgeURLActions = array(
     'deleted_post',                     // Delete a post
     'edit_post',                        // Edit a post - includes leaving comments
     'delete_attachment',                // Delete an attachment - includes re-uploading
+    'post_update',                      // Update a post
 );
 
 foreach ($cloudflarePurgeURLActions as $action) {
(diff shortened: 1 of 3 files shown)


cloudflare.loader.php
```diff
@@ -83,6 +83,7 @@
     'deleted_post',                     // Delete a post
     'edit_post',                        // Edit a post - includes leaving comments
     'delete_attachment',                // Delete an attachment - includes re-uploading
+    'post_update',                      // Update a post
```
thellimist: Could you link which hook this is?

macbookandrew: Sorry…just pushed a change to fix the typo. Here’s the hook: https://developer.wordpress.org/reference/hooks/post_updated/

macbookandrew:
@thellimist This change purges the cache for single URLs when the page/post/CPT is updated. I also updated the documentation to match the new behavior.

thellimist:
Hi @macbookandrew,

The core developers of this project has changed. @IcyApril could you notify Patryk. 

Thanks

knyttl:
Hey, would this also purge the cache on scheduled page publishing?

patryk:
Thanks!

PR #207, opened 2017-12-18
Feature/cache zone tag selective clear

lipemat:
1. Caches the Zone-tag in WP object cache to allow reuses in the tag instead of and additional request on each purge. Most useful in environments with an external object cache and scripts that run high quantity imports. 
2. Allow selectively excluding a purge cache request via a filter.  Also useful to reduce the number of requests when not needed. 

The inspiration for this came out of an issue where a site was hitting the Cloudflare rate limit when importing an ical feed.


patryk:
Closing favouring #262.

PR #208, opened 2017-12-18
Purge attachment URLs

PhatKoala:
Purge attachments and their associated resized versions on attachment update.


patryk:
Thanks!

Issue #211, opened 2018-02-07
Run purge actions when custom actions are executed


piscis:
We would like a way to run purge actions e.g. `purgeCacheEverything` and `purgeCacheByRevelantURLs` when every a specific third party action is executed.

Our use case is that we have third party plugins that modify the post state and therefore make it nessesary to clear certain URLs in cloudflare.

PR #212, opened 2018-02-07
Update cloudflare.loader.php

piscis:
Allow custom actions for purge url and purge everything actions 
Issue: #211


patryk:
Thanks!

PR #213, opened 2018-03-01
Add constants for better control of HTTP2 push

robbiet480:
Added two constants to control HTTP2 server push behavior:

`CLOUDFLARE_HTTP2_SERVER_PUSH_HEADER_SIZE` - allows setting the max header size you can push. Useful for those (like me) that know exactly what fastcgi can push through.
`CLOUDFLARE_HTTP2_SERVER_PUSH_LOG` - setting this will _enable_ the somewhat... shall we say spammy... `error_log` whenever a file is over the max header size


patryk:
Thanks!

PR #217, opened 2018-04-23
FIX: 'Warning:  count(): Parameter must be an array or an object that…

michaelmano:
@thellimist 
Fix: Moves the count which is being used on a string to the array.

### DIFF
--- a/src/WordPress/Hooks.php
+++ b/src/WordPress/Hooks.php
@@ -111,8 +111,8 @@ class Hooks
     {
         if ($this->isPluginSpecificCacheEnabled()) {
             $wpDomainList = $this->integrationAPI->getDomainList();
-            $wpDomain = $wpDomainList[0];
-            if (count($wpDomain) > 0) {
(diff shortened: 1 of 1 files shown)


nhenderson:
This MR fails a `phpunit` with PHP 7.2:

```vendor/bin/phpunit
PHPUnit 4.8.35 by Sebastian Bergmann and contributors.

....................................E.....

Time: 899 ms, Memory: 6.00MB

There was 1 error:

1) CF\Test\WordPress\WordPressAPITest::testGetDomainListReturnEmpty
count(): Parameter must be an array or an object that implements Countable

/var/www/html/Cloudflare-WordPress/src/Test/WordPress/WordPressAPITest.php:94

FAILURES!
Tests: 42, Assertions: 114, Errors: 1.
```
Please merge my merge request instead as it passes a `phpunit` on PHP 7.2:

zackproser:
Thanks so much for contributing a fix for this - I've merged the changes in #231 to address this issue.

Issue #210, opened 2018-01-16
count(): Parameter must be an array or an object that implements Countable


ahmed-sigmalux:
After upgrading to PHP 7.2, I'm seeing this repeatedly in my debug.log:

`PHP Warning:  count(): Parameter must be an array or an object that implements Countable in /var/www/wordpress/plugins/cloudflare/src/WordPress/Hooks.php on line 133`

ahmed-sigmalux:
Has anyone at Cloudflare seen this bug?

piscis:
This is what we get since the upgrade from php 7.0 to 7.2:
```
PHP Warning: count(): Parameter must be an array or an object that implements Countable in .... /plugins/cloudflare/src/WordPress/Hooks.php on line 133
```

michalhalas:
Can confirm I've been receiving same error with newest available version of plugin 3.3.2 and php version 7.0.25
`Warning: count(): Parameter must be an array or an object that implements Countable in /var/www/html/wp-content/plugins/cloudflare/src/WordPress/Hooks.php on line 133`

janvitos:
Same thing here. I noticed the issue in the NGINX error log after upgrading to PHP 7.2.

```
[error] 1140#1140: *369161 FastCGI sent in stderr: "PHP message: PHP Warning:  count(): Parameter must be an array or an object that implements Countable in /wordpress/wp-content/plugins/cloudflare/src/WordPress/Hooks.php on line 133" while reading response header from upstream, client: xxx.xxx.xxx.xxx, server: www.com, request: "POST /wordpress/wp-admin/admin-ajax.php HTTP/1.1", upstream: "fastcgi://unix:/var/run/php-fpm/php-fpm.sock:", host: "www.com", referrer: "/wordpress/wp-admin/post-new.php"
```

ron2122:
same error here


Warning: count(): Parameter must be an array or an object that implements Countable in /home/blocktoken/public_html/wp-content/plugins/cloudflare/src/WordPress/Hooks.php on line 133

Warning: Cannot modify header information - headers already sent by (output started at /home/blocktoken/public_html/wp-content/plugins/cloudflare/src/WordPress/Hooks.php:133) in /home/blocktoken/public_html/wp-includes/pluggable.php on line 1216

conorbarclay:
Still getting count() errors on PHP 7.2 here. Latest version of plugin.

janvitos:
[...]

PR #221, opened 2018-08-06
FIX: count() warning by counting $wpDomainList array instead of $wpDomain integer

janvitos:
Fix for #210 

I'm pretty sure the count() warning stems from the fact that the count() function is used to count arrays and NOT integers. But in Hooks.php, it is actually trying to count an integer, here's why.

On line 133 from Hooks.php where the warning is triggered, you have the following line of code:
```
if (count($wpDomain) <= 0) {
```
$wpDomain is actually NOT an array and is declared as a value from an array on line 132:
```
$wpDomain = $wpDomainList[0];
```
Replacing line 133 with the following code fixes the count() warning:
```
if (count($wpDomainList) <= 0) {
```

### DIFF
--- a/src/WordPress/Hooks.php
+++ b/src/WordPress/Hooks.php
@@ -130,7 +130,7 @@ class Hooks
         if ($this->isPluginSpecificCacheEnabled()) {
             $wpDomainList = $this->integrationAPI->getDomainList();
             $wpDomain = $wpDomainList[0];
-            if (count($wpDomain) <= 0) {
+            if (count($wpDomainList) <= 0) {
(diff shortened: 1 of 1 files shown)


janvitos:
Just noticed someone had already suggested the same fix, so I closed mine.

My bad.

BwanaBob:
It appears that this was addressed in Version 3.3.3 of the plugin released Mar 06, 2019
----

> 3.3.3 – 2019-3-6
> Fixed
> Bug in Hooks.php causing errors in PHP 7+

PR #226, opened 2018-10-24
Replace count() with strlen()

KenLewis:
Fix warning in PHP 7.2+ caused by using count() on a string.


zackproser:
Thanks so much for contributing a fix for this - I've merged the changes in #231 to address this issue.

PR #229, opened 2019-02-02
Turn on IPv6 by default

tambry:
WordPress works fine over IPv6, so update the default setting to reflect that.


patryk:
Thanks!

PR #249, opened 2019-10-07
New filter to allow plugins to hook for purging.

Pross:
New filter `cloudflare_purge_actions` added so plugins like page builders can add their publish actions and have the cache purged.


patryk:
Thanks, but merged #212 instead.

PR #265, opened 2020-07-01
Fix case of WordPress

jbampton:


patryk:
Thanks @jbampton !

jbampton:
Thanks @patryk 👍

PR #267, opened 2020-07-01
Fix shell script weak warning: Double quote to prevent globbing and w…

jbampton:
…ord splitting.

Checked with ShellCheck.

https://www.shellcheck.net/

https://github.com/koalaman/shellcheck/wiki/SC2086


patryk:
Thanks!

PR #286, opened 2020-09-30
Update readme for APO

garrettgalow:


garrettgalow:
still has leftover crud from past stuff...will fix

PR #300, opened 2020-10-05
composer's type=wordpress-plugin

drzraf:
Some people use [Bedrock](https://github.com/roots/bedrock/) for their setup which relies on composer for managing WP plugins.
In that case, composer/installers proceeds according to the `type` of the project.
One such example can be found [here](https://github.com/roots/soil/blob/master/composer.json#L3)

This is an alternative to using wpackagist.org


manatarms:
Thanks for the pull request!

sejoker (approved):

manatarms (approved):

manatarms:
@drzraf this was released with v3.8.4. Hope it helps.

Issue #326, opened 2020-11-17
Purge Category and Tag feed cache


janvitos:
Right now, the plugin purges the cache of various feeds, like Home Page feed and Archives feed, but it doesn't purge the cache of Category feeds or Tag feeds.

It would be really helpful if the plugin purged the Category feeds and Tag feeds cache whenever a post is published, updated, trashed, etc.

sejoker:
@janvitos totally agree, it's one of the top reported issues with APO. I'm more than happy to review and publish your PR.

janvitos:
Awesome.

Just so you know, I already applied my PR to my own fork and it is working perfectly.

sejoker:
@janvitos I can confirm this change works for tags invalidation. Unfortunately I can't make it work for category pages (cache is not purged). I also don't see cache invalidation for tags and categories when post is moved to trash or deleted. Do you have similar issues in your setup?

janvitos:
That's strange @sejoker, it seems to be working fine on my end. All Cateogies are purged at least when a post is created or updated. I have to admit I did not test for trash or deletion, but I assumed that would also work since it seemed to be working for normal taxonomy pages.

I will investigate and get back to you.

janvitos:
Hi @sejoker, as you can see from this screenshot (https://imgur.com/DSk7zza), all tags and categories are purged when a post is published or updated.

This post in particular was only assigned to a single category, so only that specific category was purged. I did not test with a post assigned to multiple categories, but it should work as well since the logic in the code was already doing that for normal category purges.

janvitos:
I confirm that purging Categories / Category feeds and Tags / Tag feeds WORKS as expected when trashing or restoring a post.

janvitos:
Any news on this @sejoker? Thanks!

sejoker:
@janvitos I was able to make it work on my test environment, changes look good, we will release changes on Monday.

PR #327, opened 2020-11-17
Purge taxonomy feed URLs

janvitos:
In reference to #326

This code will purge the taxonomy feed URLs in addition to the normal taxonomy URLs.


src/WordPress/Hooks.php
```diff
@@ -187,8 +187,10 @@ public function getPostRelatedLinks($postId)
 
             foreach ($terms as $term) {
                 $termLink = get_term_link($term);
-                if (!is_wp_error($termLink)) {
+                $termFeedLink = get_term_feed_link($term->term_id, $term->taxonomy);
```
sejoker, line 190: @janvitos will this cover both Category feeds or Tag feeds?

sejoker (approved):

src/WordPress/Hooks.php
sejoker: looking into the docs https://developer.wordpress.org/reference/functions/get_term_feed_link/ it seems cover both.

sejoker (approved):

src/WordPress/Hooks.php
janvitos: > will this cover both Category feeds and Tag feeds?

Yes it will. But I see you already got your answer :)

xtuc (approved):
Nice thanks!

manatarms (approved):

PR #332, opened 2020-12-04
Add support for configuration via environment vars

joeles:
Having a the ability to configure credentials via environment vars would be quite appreciated by advanced users.
- Storing credentials in a database can be insecure if data is exported to other environments, eg: dev. 
- Reusing credentials is not flexible if dev environments are running on different domain(s).


src/WordPress/DataStore.php
```diff
@@ -63,6 +67,10 @@ public function getHostAPIUserUniqueId()
      */
     public function getClientV4APIKey()
     {
+        if (defined('CLOUDFLARE_API_KEY')) {
```
jacobbednarz, line 70: for good measure, we probably want to also ensure the value isn't empty on the constant (in all three updates below too)

```suggestion
        if (defined('CLOUDFLARE_API_KEY') && CLOUDFLARE_API_KEY != "") {
```

jacobbednarz:
awesome stuff! after battling YAML, i managed to test this out and it works great! thank you ❤️

jacobbednarz (approved):

joeles:
Sure thing! Glad to help.

Issue #337, opened 2020-12-12
APO not clearing pagination cache when posts are created


scottbuscemi:
When a new post is added, blog pagination (ie `/page/2/` or `/page/3/`) cache is not being cleared automatically. Clearing the Cloudflare cache manually resolves this issue.

orenwolf:
👍 confirmed. We see the same issue and currently work around it with page rules.

scottbuscemi:
@orenwolf How are you configuring the Page Rules to combat this?

orenwolf:
We've created a rule for "*example.com/page/*" with an Edge Cache TTL set to 30 minutes.

This is suboptimal, as it delays the pagination caches by up to 30 minutes when changes occur, but it's much better than the default 30 days!

scottbuscemi:
As a business customer, we can only set down to 1 hour.. That's quite a long time to have the wrong pagination cached for our blog.

I'll see if I can get some Cloudflare eyeballs on this :)

orenwolf:
We’re business customers and can set to 30m, you may wish to double-check.

scottbuscemi:
My mistake - we downgraded to "Pro" so only 60m.

phansheng89:
The same issue happen for Single post Pagination, the sub page cache is not cleared automatically!

sejoker:
FYI we will try to get the fix out early next week.

scottbuscemi:
Thanks @sejoker - looking forward to it.

thanks2music:
@sejoker Thanks for the great WordPress plugin.

I have a similar problem besides pagination.

1. new posts on the top page are not updated when viewed with a real browser on a smartphone (no problem with desktop browsers)
2. the category and tag pages to which the article belongs are not updated anymore.
3. when I update a published article with an addition, the actual browser on my smartphone does not update as in 1.

I am having trouble with the following three points.

I'm sorry to bother you, but I would appreciate it if you could check this for me.
Thank you very much.

garrettgalow:
Reopening this since we reverted the fix as it caused performance regressions

fidividi32:
Hi guys any update on this?
After weeks of investigating, we realized our issue is precisely this one, and our WordPress APO enabled website is sitting on CloudFlare with this issue causing angry customer/users due to old posts appearing on subsequent category pages. Thanks for any updates.

sejoker:
[...]

PR #339, opened 2020-12-19
add pagination purging

sejoker:
Fixes https://github.com/cloudflare/Cloudflare-WordPress/issues/337

### DIFF
--- a/src/WordPress/Hooks.php
+++ b/src/WordPress/Hooks.php
@@ -234,6 +234,20 @@ class Hooks
 
         // Home Page and (if used) posts page
         array_push($listofurls, home_url('/'));
+
+        // Refresh pagination
+        $total_posts_count = wp_count_posts()->publish;
+        $posts_per_page = get_option('posts_per_page');
+        $page_number_max = ceil($total_posts_count / $posts_per_page);
+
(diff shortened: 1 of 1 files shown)


xtuc (approved):

scottbuscemi:
@sejoker This seems to cover /page/*, but are /category-name/* and /tag-name/* pagination covered as well?

SmartandCom:
Pagination problem solved, but it looks like it'as very heavy process in the background, updating or publishing something takes a very long time with last version of the plugin, like if purging is working for ever

scottbuscemi:
> Pagination problem solved, but it looks like it's a very heavy process in the background, updating or publishing something takes a very long time with last version of the plugin, like if purging is working for ever

How many posts do you have? I was afraid this solution would slow things down.

SmartandCom:
7600 posts paginated within 1530 pages
indeed it looks like the cache clearing process is taking a ,lot of time. Our server is powerful enough not  to hang and go to 500 error but it still takes  minutes to process.

xtuc:
We have a tracking issue https://github.com/cloudflare/Cloudflare-WordPress/issues/351 and we are planning to revert the change.

PR #350, opened 2021-01-14
Get CI running again

jacobbednarz:
CI was failing to run all tasks due to some underlying dependency issues. The 
biggest of which was that we were using Composer 2 and not all of the plugins 
were compatible. To remedy this, I've swapped the steps to 
`php-actions/composer@v5` which allows us to specify which version in 
combination with the PHP version.


phpcs.xml
```diff
@@ -6,4 +6,8 @@
 
     <rule ref="PHPCompatibility"/>
     <rule ref="PSR2"/>
+
+    <rule ref="PSR2">
+      <exclude name="Generic.Files.LineLength"/>
```
jacobbednarz, line 11: I've opted to ignore long lines as it doesn't look like it's really been enforced previously.

jacobbednarz:
Do we still need the Travis CI run here? Or can we remove that?

patryk:
Yeah, I vote get rid of Travis-CI :)

patryk (approved):

jacobbednarz:
have removed travis configuration in 19d9a44 and disabled the webhook in the repository settings.

manatarms (approved):

PR #356, opened 2021-01-19
Change purge logic for comments

softmanro:
The existing implementation of the Cloudflare plugin triggers a cache purge whenever a new comment is added, no matter if the comment might be held for moderation. My PR proposes a change in order to make cache purging "aware" of comment status and only act (e.g. purge cache) upon significant/visible changes with regard to the post.

### DIFF
--- a/src/WordPress/Hooks.php
+++ b/src/WordPress/Hooks.php
@@ -325,24 +325,33 @@ class Hooks
 
     public function purgeCacheOnCommentStatusChange($new_status, $old_status, $comment)
     {
-      if (!is_object($comment)) return; // nothing to do
-      if (empty($comment->comment_post_ID)) return; // nothing to do
+        if (!is_object($comment)) {
+            return; // nothing to do
+        }
+        if (empty($comment->comment_post_ID)) {
+            return; // nothing to do
+        }
 
       // in case the comment status changed, and either old or new status is "approved", we need to purge cache for the corresponding post
-      if (($old_status != $new_status) && (($old_status === 'approved') || ($new_status === 'approved'))) {
-        $this->purgeCacheByRelevantURLs($comment->comment_post_ID);
-        return;
-      }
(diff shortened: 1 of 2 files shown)


sejoker:
Thanks for the proposed change, it does make sense, I will need to test it more before approving though.

softmanro:
Thank you for the update. I've performed some testing on various scenarios (moderation on or off, new comment, approve/unapprove, trash unapproved, trash approved & so on) but of course such a change needs enough testing in order not to break other's setups. Mine is live and running with these changes for about a week and I haven't seen unexpected issues.
Looking forward to your feedback once testing is complete.

jacobbednarz:
@sejoker this looks 👌 to me. do you have any reservations about merging following your testing? if not, i'm happy to 🚢

src/WordPress/Hooks.php
```diff
@@ -322,4 +322,36 @@ public function initAutomaticPlatformOptimization()
...
     }
+
+    public function purgeCacheOnCommentStatusChange($new_status, $old_status, $comment)
+    {
+        if (!is_object($comment)) {
+            return; // nothing to do
+        }
+        if (empty($comment->comment_post_ID)) {
+            return; // nothing to do
+        }
```
jacobbednarz, line 333: we can opt to simplify this one a little with `!isset` and `empty` checks in the same expression

```suggestion
        if (!isset($comment->comment_post_ID) || empty($comment->comment_post_ID)) {
            return;  
        }
```

softmanro: That is right, Jacob - thank you.

sejoker:
@jacobbednarz  please go ahead with merging and publishing.

sejoker:
this change should also decrease the number of duplicated purges as we don't need to subscribe to `edit_post` event. Thanks @softmanro for your PR.

sejoker (approved):

jacobbednarz:
thank you both! 👏

PR #363, opened 2021-01-28
Drop PHP 5.6, 7.0 and 7.1 support

jacobbednarz:
Note: This PR started out as one thing, however, due to finding some interlocking
dependency issues, it became a much larger thing.

This PR drops support for PHP 5.6, 7.0 and 7.1. The issue started off as a failed 
SVN commit to [release a new version](https://github.com/cloudflare/Cloudflare-WordPress/runs/1787941339?check_suite_focus=true).

```
svn: E165001: Commit failed (details follow):
svn: E165001: Commit blocked by pre-commit hook (exit code 1) with output:

***********************************
PHP error in: cloudflare/tags/3.8.10/vendor/sebastian/diff/tests/ParserTest.php:

Parse error: Invalid body indentation level (expecting an indentation level of at least 1) in Standard input code on line 78
Errors parsing Standard input code
***********************************
```

Despite being a vendored dependency, it is still checked against `php -l` to as
a safeguard to prevent bad releases from impacting end users. Per the CI output,
the check is run against 7.3 however the issue is also present in 7.2.

```
$ PHP_VERSION=7.2.17 php -l vendor/sebastian/diff/tests/ParserTest.php

Parse error: Invalid body indentation level (expecting an indentation level of at least 1) in vendor/sebastian/diff/tests/ParserTest.php on line 78
Errors parsing vendor/sebastian/diff/tests/ParserTest.php
```

This unfortunately leads down a bit of a rabbit hole as dependencies are now 
involved and we were previously straddling the PHP 5 and PHP 7 border while
still supporting versions of PHP that neither [WordPress](https://en-au.wordpress.org/about/requirements/) or [PHP](https://www.php.net/supported-versions.php) are willing to
support anymore. 

To rememdy, we've decided to drop support for PHP 5.6, 7.0 and 7.1 which allows 
us to update the dependencies as well as focus our efforts on more modern 
versions without needing to backport fixes or support custom forks. The versions
we have opted to drop are well out of date and while it may interrupt some 
workflows, the rest of the ecosystem is moving forward and we need to get a move 
on too.

WordPress's `readme.txt` allows to define the [minimum PHP version](https://make.wordpress.org/plugins/2017/08/29/minimum-php-version-requirement/) which _should_ 
prevent any breakages for future updates.

A side effect of updating these dependencies is that `phpunit/php-code-coverage` 
has a file within it's test suite that intentionall causes a crash. While a 
helpful test, it's going to trip up our CI checks and SVN pushes (similar the 
original issue which sparked this) so I've added a custom composer command to 
remove it from the vendor directory on CI runs and release steps. Unless someone
attempts to run the vendor specific test suite, this will be a noop.


manatarms (approved):

.github/workflows/php.yml
```diff
@@ -14,18 +14,21 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php-versions: ['5.6', '7.0', '7.1', '7.2', '7.3', '7.4']
+        php-versions: ['7.2', '7.3', '7.4']
 
     steps:
     - uses: actions/checkout@v2
 
+    - uses: nanasess/setup-php@master
```
patryk, line 22: Did you consider https://github.com/shivammathur/setup-php instead of this?

composer.json
```diff
@@ -25,15 +25,16 @@
     },
     "scripts": {
         "format": "php vendor/bin/phpcs -d date.timezone=UTC --standard=phpcs.xml",
-        "test": "php vendor/bin/phpunit"
+        "test": "php vendor/bin/phpunit",
```
patryk, line 28: Should we exclude vendor directory here (or `phpunit.xml`)?

.github/workflows/php.yml
jacobbednarz: didn't even know it existed 😆 i just got this one from Googling so happy to swap if you have a preference.

composer.json
jacobbednarz: is https://github.com/cloudflare/Cloudflare-WordPress/blob/master/phpunit.xml#L5 what you mean? if so, we've already got that handled (same for phpcs too - https://github.com/cloudflare/Cloudflare-WordPress/blob/master/phpcs.xml#L3).

.github/workflows/php.yml
patryk: Yeah, I'd recommend try it out. That user has many more stars and other PHP-related projects.

jacobbednarz: sure, swapped in 0465c9c

PR #365, opened 2021-02-09
Purge posts that go from public to private.

Inwerpsel:
Currently if a post was published publicly and then gets set to private, no cache purge is done, even though that's exactly what you would expect to happen in that case. This is because the hook returns early unless the post is "publish" or "trash" status. I changed it so that if a post is private but it was not private previously (i.e. when the private status is being applied), it still executes the purge.

* Add ...$args to the hook to avoid doing additional queries. This could
likely be used elsewhere in the function instead of querying those
things.
* Make "private" also a "valid" post status, add extra check to still
skip purge unless it's a status change.

### DIFF
--- a/src/WordPress/Hooks.php
+++ b/src/WordPress/Hooks.php
@@ -126,7 +126,7 @@ class Hooks
         }
     }
 
-    public function purgeCacheByRelevantURLs($postId)
+    public function purgeCacheByRelevantURLs($postId, ...$args)
     {
         if ($this->isPluginSpecificCacheEnabled() || $this->isAutomaticPlatformOptimizationEnabled()) {
             $wpDomainList = $this->integrationAPI->getDomainList();
@@ -135,11 +135,28 @@ class Hooks
             }
             $wpDomain = $wpDomainList[0];
 
-            $validPostStatus = array('publish', 'trash');
+            $validPostStatus = array('publish', 'trash', 'private');
             $thisPostStatus = get_post_status($postId);
 
             if (get_permalink($postId) != true || !in_array($thisPostStatus, $validPostStatus)) {
                 return;
+            }
+
+            // We don't need to purge for private posts, except when they were just transitioned from public.
+            if ('private' === $thisPostStatus) {
+                // The action that fires on transition of status is "post_updated", which will have the old version
+                // in the extra args. If that arg is not there we can bail out.
+                if (! isset($args[1])) {
+                    return;
+                }
+                /**
+                 * @var \WP_Post $before Previous version of post.
+                 */
+                list(, $before) = $args;
+
+                if ('private' === $before->post_status) {
+                    return;
+                }
             }
 
             if (is_int(wp_is_post_autosave($postId)) ||  is_int(wp_is_post_revision($postId))) {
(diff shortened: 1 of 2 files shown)


src/WordPress/Hooks.php
```diff
@@ -126,7 +126,7 @@ public function purgeCacheEverything()
         }
     }
 
-    public function purgeCacheByRelevantURLs($postId)
+    public function purgeCacheByRelevantURLs($postId, ...$args)
```
jacobbednarz, line 129: @Inwerpsel i'm fresh to this codebase so excuse the potentially stupid question but are you able to clarify for me why we're updating the function arguments? from what i can see, i think we can do it within the method itself without modifying this and causing a potentially breaking change for the plugin.

sejoker: this the explanation from the PR description:
```
Add ...$args to the hook to avoid doing additional queries. This could
likely be used elsewhere in the function instead of querying those
things.
```
I prefer to play safe with a potentially breaking change.

sejoker: To be fair I think it is ok change. @Inwerpsel  is filtering out the case when post was updated and  state wasn't changed (private -> publish). We only interested in (publish -> private). For that we need 3rd parameter https://developer.wordpress.org/reference/hooks/post_updated/

src/WordPress/Hooks.php
```diff
@@ -135,13 +135,30 @@ public function purgeCacheByRelevantURLs($postId)
...
+                // in the extra args. If that arg is not there we can bail out.
+                if (! isset($args[1])) {
+                    return;
+                }
+                /**
+                 * @var \WP_Post $before Previous version of post.
+                 */
+                list(, $before) = $args;
+
+                if ('private' === $before->post_status) {
```
sejoker, line 157: We should change to `$before->post_status !== 'private'` as we don't interested in `trash` status or any else.

sejoker (approved):
I've tested the changes, works ok.

sejoker:
@Inwerpsel thanks for contribution.

src/WordPress/Hooks.php
Inwerpsel: Hey @sejoker ! Sorry for losing track of this PR and replying this late, so I have to dig in my memory a bit. I think I did it to be more future proof, not sure if it's already needed in some cases. If a post transitions to the `private` status, it should definitely not be in the cache. I can image there could be cases where it somehow does not directly transition from `publish` to `private`, but it was still previously cached.

src/WordPress/Hooks.php
Inwerpsel: @jacobbednarz Updating the function arguments should by itself be safe for the actions that are listed, since the code is already written to support running on multiple actions, each of which can have a different number of arguments.

But maybe it does have some implications if custom actions are added with `cloudflare_purge_url_actions` :thinking: These could have any function signature, while these checks make some assumptions about the presence of the 3rd argument meaning it's a `post_updated` action.

https://github.com/greenpeace/Cloudflare-WordPress/blob/d7784bf47f79dcf3e4f758df5e7f75c696cfc203/src/WordPress/Hooks.php#L149-L155

PR #374, opened 2021-02-16
Fix #247 - unneeded symfony/yaml dependency

### LINKED ISSUE #247: Unneeded Symfony/Yaml dependency? (and bundled dev dependencies)
I've been running into some problems with a wp-cli command that needs a newer version of Symfony/Yaml than is vendored with this plugin, and this plugin ends up blocking it from being loaded because of autoloader precedence.

However, I can't seem to find where the Yaml component is actually *used* by anything within the Cloudflare plugin, so the dependency appears redundant.  (In addition, the bundled vendor/ directory appears to have been built with development dependencies, meaning the distribution is unnecessarily large and prone to overriding other packages with outdated versions as well.)

Perhaps symfony/yaml can now be removed from composer.json and the vendor tree rebuilt with just the non-development dependencies?  Thanks.

pjeby:


jacobbednarz:
Do you have composer V1? You can fetch it and install using the following

```

php -r "copy('https://getcomposer.org/installer', 'composer-setup.php');"

php -r "if (hash_file('sha384', 'composer-setup.php') === '756890a4488ce9024fc62c56153228907f1545c228516cbf63f885e036d37e9a59d27d63f46af1d4d07ee0f76181c7d3') { echo 'Installer verified'; } else { echo 'Installer corrupt'; unlink('composer-setup.php'); } echo PHP_EOL;"

php composer-setup.php --version=1.10.20 --filename=composer-v1

php -r "unlink('composer-setup.php');"
```

Then you can update with `php composer-v1 update`

pjeby:
I ended up just editing the composer.lock API version back to what it was before so it doesn't break your processes, after verifying that the only change of substance to the .lock file was moving yaml and ctypes from packages to packages-dev.

jacobbednarz (approved):
this looks good to me and still works fine with local testing, thank you! 🍰

PR #383, opened 2021-03-07
read root level composer + user configuration files

### LINKED ISSUE #380: Bug: Warning: file_get_contents of composer.json
Latest version of CF plugin, when visiting WP admin Settings->Cloudflare dashboard.. You can see a red bar below in a page, and getting an error each time:

```
wp-content/plugins/cloudflare/src/WordPress/composer.json): failed to open stream: No such file or directory
```

jacobbednarz:
During the CI revamp, we pulled out a manual step that would copy the
`composer.json` file into the same directory as the `PluginsActions`
class. This seems quite verbose so instead of restoring that
functionality, let's just reuse the root level files.

Closes #380

### DIFF
--- a/src/WordPress/PluginActions.php
+++ b/src/WordPress/PluginActions.php
@@ -53,6 +53,9 @@ class PluginActions extends AbstractPluginActions
         'integrationName',
     ];
 
+    const USER_CONFIG_PATH = '/../../config.json';
+    const COMPOSER_CONFIG_PATH = '/../../composer.json';
+
     public function __construct(DefaultIntegration $defaultIntegration, APIInterface $api, Request $request)
     {
         parent::__construct($defaultIntegration, $api, $request);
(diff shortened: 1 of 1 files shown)


manatarms (approved):

src/WordPress/PluginActions.php
```diff
@@ -53,6 +53,9 @@ class PluginActions extends AbstractPluginActions
         'integrationName',
     ];
 
+    const USER_CONFIG_PATH = '/../../config.json';
```
manatarms, line 56: I remember at some point there was a script that nuked the root level config and composer jsons. Can we confirm that these exist in the release zips?

jacobbednarz: yep, we definitely do keep it around. both archives (GitHub + WordPress SVN) are showing it as present.

```
$ wget https://downloads.wordpress.org/plugin/cloudflare.4.2.1.zip
$ unzip -l cloudflare.4.2.1.zip | grep -e "composer\.json" -e "config\.json"
     1309  03-05-2021 20:13   cloudflare/composer.json
      807  03-05-2021 20:13   cloudflare/config.json
      879  12-12-2017 15:07   cloudflare/vendor/cloudflare/cf-ip-rewrite/composer.json
      594  10-02-2020 04:01   cloudflare/vendor/cloudflare/cloudflare-plugin-backend/composer.json
      561  06-29-2020 13:39   cloudflare/vendor/psr/log/composer.json
```

PR #390, opened 2021-03-17
Unpin dev dependencies, add PHP 8 support

jacobbednarz:
This PR started off as adding PHP 8.0 to the build matrix in order to test what our upgrade position will be once we decided to roll to PHP 8.0 however upon running an initial test, the only failures were a couple of tests so I decided to unpin the dependencies at the same time. By doing this, we're no longer running out of date development dependencies with a weird versioning web of constraints; we're running the latest stable available.

The changes here are `development` only (dev-deps and tests) so there shouldn't be any end user changes expected.

### DIFF
--- a/.github/workflows/php.yml
+++ b/.github/workflows/php.yml
@@ -10,11 +10,16 @@ jobs:
   build:
 
     runs-on: ubuntu-latest
-
+    continue-on-error: ${{ matrix.experimental }}
     strategy:
       fail-fast: false
       matrix:
         php-versions: ['7.2', '7.3', '7.4']
+        experimental: [false]
+        include:
+          - php-versions: '8.0'
+            experimental: true
+
 
     steps:
     - uses: actions/checkout@v2
--- a/src/WordPress/PluginActions.php
+++ b/src/WordPress/PluginActions.php
@@ -87,7 +87,7 @@ class PluginActions extends AbstractPluginActions
             throw new ZoneSettingFailException();
         }
 
-        $currentPlan = $details['result']['plan']['legacy_id'];
+        $currentPlan = $details['result']['plan']['legacy_id'] ?? 'free';
 
         $result &= $this->clientAPI->changeZoneSettings($zoneId, 'security_level', array('value' => 'medium'));
         if (!$result) {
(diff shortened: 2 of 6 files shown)


src/WordPress/PluginActions.php
```diff
@@ -87,7 +87,7 @@ public function applyDefaultSettings()
             throw new ZoneSettingFailException();
         }
 
-        $currentPlan = $details['result']['plan']['legacy_id'];
+        $currentPlan = $details['result']['plan']['legacy_id'] ?? 'free';
```
jacobbednarz, line 90: this is the result of fixing 

```
1) CF\WordPress\Test\PluginActionsTest::testReturnApplyDefaultSettingsChangeZoneSettingsThrowsZoneSettingFailException
Trying to access array offset on value of type bool

/app/src/WordPress/PluginActions.php:90
/app/src/Test/WordPress/PluginActionsTest.php:121
```

Prior to this change, it was returning `false` (as I assume one of the keys weren't present) however I updated it to default to "free" which should cover all of our use cases and a safe assumption given free accounts can do basic things.

src/WordPress/Proxy.php
```diff
@@ -89,7 +89,7 @@ public function createRequest()
                 $path = API\Plugin::ENDPOINT . $_GET['proxyURL'];
             }
         } else {
-            $path = $body['proxyURL'];
+            $path = $body['proxyURL'] ?? '';
```
jacobbednarz, line 92: similar to the above, this was a failure in CI

```
2) CF\Test\WordPress\ProxyTest::testRunHandlesPost
Trying to access array offset on value of type null

/app/src/WordPress/Proxy.php:92
/app/src/WordPress/Proxy.php:58
/app/src/Test/WordPress/ProxyTest.php:77
```

Again, it was returning `null` so instead I defaulted to an empty `proxyURL` value which _may_ still be desirable. If not, we need to rejig the test case.

.github/workflows/php.yml
```diff
@@ -10,11 +10,16 @@ jobs:
...
     runs-on: ubuntu-latest
-
+    continue-on-error: ${{ matrix.experimental }}
     strategy:
       fail-fast: false
       matrix:
         php-versions: ['7.2', '7.3', '7.4']
+        experimental: [false]
+        include:
+          - php-versions: '8.0'
```
jacobbednarz, line 20: adds PHP 8.0 but allows it to fail if a new release comes out and we're not tracking it too closely.

mgirouard (approved):

deuill (approved):

PR #391, opened 2021-03-18
Purge pagination

sejoker:
Second try to fix https://github.com/cloudflare/Cloudflare-WordPress/issues/337. In addition to previous https://github.com/cloudflare/Cloudflare-WordPress/pull/339/files I limit the number of pages to 3.
Also APO applies Edge Cache TTL of 30 mins for all `example.com/page/*` paginated urls.

### DIFF
--- a/src/WordPress/Hooks.php
+++ b/src/WordPress/Hooks.php
@@ -126,7 +126,7 @@ class Hooks
         }
     }
 
-    public function purgeCacheByRelevantURLs($postId)
+    public function purgeCacheByRelevantURLs($postId, ...$args)
     {
         if ($this->isPluginSpecificCacheEnabled() || $this->isAutomaticPlatformOptimizationEnabled()) {
             $wpDomainList = $this->integrationAPI->getDomainList();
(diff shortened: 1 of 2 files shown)


src/WordPress/Hooks.php
```diff
@@ -256,6 +256,20 @@ public function getPostRelatedLinks($postId)
             array_push($listofurls, $pageLink);
         }
 
+        // Refresh pagination
+        $total_posts_count = wp_count_posts()->publish;
+        $posts_per_page = get_option('posts_per_page');
+        // Limit to up to 3 pages
+        $page_number_max = min(3, ceil($total_posts_count / $posts_per_page));
```
xtuc, line 263:
```suggestion
        $page_number_max = max(3, ceil($total_posts_count / $posts_per_page));
```

xtuc (approved):

src/WordPress/Hooks.php
jacobbednarz, line 263: i wonder if we should make this configurable for those who want to do say 2 or 10 instead of the static 3?

jacobbednarz (approved):
A non-blocking question, otherwise 🚢

src/WordPress/Hooks.php
```diff
@@ -256,6 +256,20 @@ public function getPostRelatedLinks($postId)
...
+
+        $this->logger->debug("total_posts_count $total_posts_count");
```
jacobbednarz, line 265: do we need to keep these debug lines around?

sejoker: > do we need to keep these debug lines around?

I think yes, we don't enable logging by default, so when we need logging it's useful to include.

src/WordPress/Hooks.php
sejoker: We can do this once requested by customers. I think it's ok to ship without configuration option.

PR #396, opened 2021-03-21
Purge a post when clean_post_cache is fired on it

### LINKED ISSUE #395: APO does not purge when a scheduled post is published
The `post_updated` hook is insufficient to clear a post from the cache, because it does not fire on all changes to a post. When a post is scheduled, APO won't purge it, and thus it won't appear on the site until manually re-saved.

`clean_post_cache` would be a good addition to the list, as this is what WordPress internally uses to purge a post from its object cache.

markjaquith:
fixes #395

If you think this is too aggressive, then another solution would be to listen on `transition_post_status` and purge whenever a post transitions to or from `publish` (including publish-to-publish).

### DIFF
--- a/cloudflare.loader.php
+++ b/cloudflare.loader.php
@@ -96,6 +96,7 @@ $cloudflarePurgeURLActions = array(
     'deleted_post',                     // Delete a post
     'delete_attachment',                // Delete an attachment - includes re-uploading
     'post_updated',                     // Update a post
+    'clean_post_cache',                 // Clean a post from the WordPress object cache
 );
 
 $cloudflarePurgeURLActions = apply_filters('cloudflare_purge_url_actions', $cloudflarePurgeURLActions);


jacobbednarz:
By too aggressive, do you mean we want to be more selective about what post statuses we attempt to clear the cache on? If so, I think I agree we want to refine this to posts where the publish state is the source or destination to avoid unnecessarily purging. 

@sejoker do you have a preference?

markjaquith:
Yes, that's what I mean. WordPress will clear the post cache even on draft-to-draft saves.

I'll open a new PR that targets post status transitions more surgically. I think the correct approach to listening to `transition_post_status` is:

`$oldStatus === 'publish' || $newStatus === 'publish'`

That'll handle:

- publishing a post
- unpublishing a post
- updating a published post
- a scheduled post being published automatically by WordPress

markjaquith:
Closing in favor of #397

PR #397, opened 2021-03-21
Purge a post from APO when its status transitions to or from "publish"

### LINKED ISSUE #395: APO does not purge when a scheduled post is published

markjaquith:
This includes publish => publish transitions (editing a published post), manually publishing a post, unpublishing a post, or WordPress automatically publishing a scheduled post at the appropriate time.

I removed the `post_updated` hook because it would be redundant, and the current architecture of the plugin would result in multiple purge requests being sent for the same post update.

see #396 (different approach)
fixes #395

### DIFF
--- a/cloudflare.loader.php
+++ b/cloudflare.loader.php
@@ -100,7 +100,7 @@ $cloudflarePurgeURLActions = array(
 $cloudflarePurgeURLActions = apply_filters('cloudflare_purge_url_actions', $cloudflarePurgeURLActions);
 
 foreach ($cloudflarePurgeURLActions as $action) {
-    add_action($action, array($cloudflareHooks, 'purgeCacheByRelevantURLs'), PHP_INT_MAX, 4);
+    add_action($action, array($cloudflareHooks, 'purgeCacheByRelevantURLs'), PHP_INT_MAX);
 }
 
 /**
--- a/cloudflare.loader.php
+++ b/cloudflare.loader.php
@@ -95,7 +95,6 @@ foreach ($cloudflarePurgeEverythingActions as $action) {
 $cloudflarePurgeURLActions = array(
     'deleted_post',                     // Delete a post
     'delete_attachment',                // Delete an attachment - includes re-uploading
-    'post_updated',                     // Update a post
 );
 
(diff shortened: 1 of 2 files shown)


sejoker (approved):

sejoker:
@markjaquith, thanks for contributing. I initially approved your PR but realized we need to fuse it with the recent https://github.com/cloudflare/Cloudflare-WordPress/pull/365/files. It was fixing a related use case. I think your change supersedes it, so we need to revert https://github.com/cloudflare/Cloudflare-WordPress/pull/365/files as part of your PR.

sejoker:
Tagging @Inwerpsel just in case you have suggestions around this PR.

markjaquith:
@sejoker Yeah, that assumption of what the additional args are going to be for a variety of different hooks is not a great idea, IMO.

There's another issue that's related, which is that purges are fired immediately, so you can't collect "to-purge" URLs from a bunch of different potentially firing hooks and then deduplicate them before sending them off.

And related to **that** is the issue of synchronous purges during save events, which, although great for "purges happen quickly", slows down the publishing experience and can lead to local Varnish caches being purged **after** APO, leading to "stuck" content, but that's another issue for another day. 😄

sejoker (approved):

sejoker:
@markjaquith  good points, I will need to test it a bit more on my WordPress setup, but LGTM in terms of code changes.

jacobbednarz:
This looks great; good pick up on the conflicting PR here @sejoker now that we have a more robust option.

Thanks for this @markjaquith! We'll aim to include it in the next release which is due today or tomorrow.

Inwerpsel:
@sejoker Indeed looks like a simpler way to address the same issue. :ok_hand: [I created a PR](https://github.com/cloudflare/Cloudflare-WordPress/pull/403) to add the description of the behavior of `transition_post_status` this relies upon to the code itself as a comment. I first saw the diff before I saw the PR description, and it's not intuitive how the current hook also accounts for purges when the status doesn't actually transition.

PR #403, opened 2021-04-07
Document unintuitive WP hook behavior

Inwerpsel:
* The `transition_post_status` hook is actually misnamed, since it also
runs when the status doesn't change but the post is edited. This plugin
relies on this behavior to purge published posts on changes. I copied
and slightly authored the documentation for this that was only present
in [the PR description](https://github.com/cloudflare/Cloudflare-WordPress/pull/397#issue-597536270).

### DIFF
--- a/cloudflare.loader.php
+++ b/cloudflare.loader.php
@@ -105,6 +105,10 @@ foreach ($cloudflarePurgeURLActions as $action) {
 
 /**
  * Register action to account for post status changes
+ * This includes
+ * - publish => publish transitions (editing a published post: no actual status change but the hook runs nevertheless)
+ * - manually publishing/unpublishing a post
+ * - WordPress automatically publishing a scheduled post at the appropriate time
  */
 add_action('transition_post_status', array($cloudflareHooks, 'purgeCacheOnPostStatusChange'), PHP_INT_MAX, 3);
 


sejoker (approved):

sejoker:
LGTM.

jacobbednarz (approved):
I'm generally against code documentation like this as we should be endeavouring to write code that doesn't need explanation and only commenting where the "how" is not clear but I take your observation onboard that the hook is unintuitive so 👍 to adding this one for clarity.

lkraav:
You should probably add WP version information in such comments, and/or tagged tree code links in `@see` tag, for maximum clarity.

These things are subject to change and people would know better to check for upstream updates.

markjaquith:
Looks good.

FWIW, although this hook is poorly named, it has been this way since 2007, and this behavior (firing on post update where the status does not change) is documented. It ain't changing. 😄

jacobbednarz:
@lkraav i agree with @markjaquith here; it seems excessive to duplicate the documentation into this repository too.

@Inwerpsel what is stopping us from transitioning this back to ready to merge?

lkraav:
> duplicate the documentation

Who suggested duplicate documentation?

I do `@see https://core.trac.wordpress.org/browser/tags/5.7.1/src/wp-includes/functions.wp-scripts.php#L343` type annotations, where version tag is usually the latest available (applicable to the specific problem).

Issue #404, opened 2021-04-09
APO cache is not cleared properly.


rvola:
### Confirmation

My issue isn't already found on the issue tracker.
I have replicated my issue using the latest version of the plugin and it is still present.

### WordPress version

5.7

### Cloudflare-WordPress version

4.4.0

### PHP version

7.4.0

### Expected result

that the APO cache after a change / publish / etc really clears

### Actual result

Only the desktop version cache is cleared

### Steps to reproduce

Activate the plugin
Activate APO
Activate "Cache by device type"
Create content.
View the page with mobile and desktop (the order does not matter).
Publish new content.
Top up on mobile (nothing has changed: no new content)
Reload on desktop (new content appears)

### Additional factoids

I investigated the problem at length and here are my conclusions.

When the "[zonePurgeFiles](https://github.com/cloudflare/Cloudflare-WordPress/blob/f87045b0227de24e57af4d4d88d88583afa0468d/src/WordPress/WordPressClientAPI.php#L59)" method is executed with an array of 30 lines, it clears the cache well but not in mobile (tablet I have not checked).

I modified the request by adding the headers "CF-Device-Type" but nothing changes.

If I delete the url in the Cloudflare interface directly it clears all the caches on that page.

If I modify the request not to send an array but just 1 url like this (imagining that the first line $files[0] is my test url 😉):

```
$request = new Request('DELETE', 'zones/'.$zoneId.'/purge_cache', array(), array('files' => $files[0]));
$response = $this->callAPI($request);
```
the cache clears well.

Looks like the request by array does not clear the devicetype cache

rvola:
similary #394

rvola:
here is some additional feedback.

I need to use mobile / desktop cache unbundling despite my site being responsive, there is an ad script that can't work properly.

I also use before activating APO the plugin WP-ROCKET with options:

- [x] Enable caching for mobile devices
- [x] Separate cache files for mobile devices

I did a test:

1. Clear all WP-Rocket cache
2. Clear all Cloudflare cache
3. Load the home on the desktop (WP-ROCKET OK / Cloudflare OK file)
4. Load the home on the mobile (WP-ROCKET OK / Cloudflare OK file)
5. Creation of a new article
6. WP Rocket erases index-https.html (with gzip too) and index-mobile-https.html (with gzip too)
7. Loading the page on desktop: OK Cloudlfare updated with the 'age' header
[...]

PR #405, opened 2021-04-09
Small fix on the list of urls to clear the cache

rvola:
Hello everyone 👋
by trying to correct the problem mentioned here: #404 
I found 2 small improvements:

### Do not send private taxonomy - 1488f65
I noticed that if the user has a private taxonomy, we still send in the table of urls) delete urls to Cloudflare on not-public taxonomy.
In itself it is not annoying, but in the case of many urls, we limit calls to Cloudflare by being more efficient.

### Check that the table is full - eed1052
Likewise it can happen that by the specific construction of custom taxonomy we refuse the feed urls.
Line [192](https://github.com/cloudflare/Cloudflare-WordPress/blob/64fcc91f53d8599b5ca80b3f4430c9f61d8cdd01/src/WordPress/Hooks.php#L192) empty referral.
I added a small row that will clean up the empty rows in the table.

Thank you

### DIFF
--- a/src/WordPress/Hooks.php
+++ b/src/WordPress/Hooks.php
@@ -174,7 +174,7 @@ class Hooks
         $postTypeTaxonomies = get_object_taxonomies($postType);
 
         foreach ($postTypeTaxonomies as $taxonomy) {
-            //Only if taxonomy is public
+            // Only if taxonomy is public
             $taxonomy_data = get_taxonomy($taxonomy);
             if ($taxonomy_data instanceof WP_Taxonomy && false === $taxonomy_data->public) {
                 continue;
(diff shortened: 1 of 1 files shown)


rvola:
sorry for the force-push, there was a syntax problem and I don't want to pollute a commit.

sejoker:
LGTM.

sejoker (approved):

src/WordPress/Hooks.php
```diff
@@ -173,6 +174,12 @@ public function getPostRelatedLinks($postId)
         $postTypeTaxonomies = get_object_taxonomies($postType);
 
         foreach ($postTypeTaxonomies as $taxonomy) {
+            //Only if taxonomy is public
```
jacobbednarz, line 177:
```suggestion
            // Only if taxonomy is public
```

jacobbednarz (approved):

jacobbednarz:
thanks for this @rvola! 🌟

rvola:
Thanks you also for your plugin 👍

PR #410, opened 2021-05-19
Only add attachment URL to purge list if WordPress provides info about it

markjaquith:
`wp_get_attachment_image_src()` can return `false` for a given postId/size combo, in which case we should skip past it.

The lack of this test is causing this:

`Trying to access array offset on value of type bool`

CleanShot 2021-05-19 at 12 32 38@2x

### DIFF
--- a/src/WordPress/Hooks.php
+++ b/src/WordPress/Hooks.php
@@ -259,7 +259,7 @@ class Hooks
             $attachmentUrls = array();
             foreach (get_intermediate_image_sizes() as $size) {
                 $attachmentSrc = wp_get_attachment_image_src($postId, $size);
-                if ($attachmentSrc) {
+                if (is_array($attachmentSrc) && !empty($attachmentSrc)) {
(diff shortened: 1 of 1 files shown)


src/WordPress/Hooks.php
```diff
@@ -259,7 +259,9 @@ public function getPostRelatedLinks($postId)
             $attachmentUrls = array();
             foreach (get_intermediate_image_sizes() as $size) {
                 $attachmentSrc = wp_get_attachment_image_src($postId, $size);
-                $attachmentUrls[] = $attachmentSrc[0];
+                if ($attachmentSrc) {
```
jacobbednarz, line 262: let's be a little more explicit here as the return type is either array or boolean. it's not immediately clear from the code that `$attachmentSrc` can be either and it may throw some people off. 

```suggestion
                if (is_array($attachmentSrc) && !empty($attachmentSrc)) {
```

jacobbednarz (approved):

Issue #345, opened 2021-01-05
Hooks called to early for themes


marcwieland95:
Hey there

I think a couple of hooks are called too early to be able to alter them via a theme.

`cloudflare_purge_everything_actions` and `cloudflare_purge_url_actions` should be called within an `init` action.

Can someone proof that?
Thanks and cheers
Marc

markjaquith:
Correct... it's bad practice to apply filters or do actions as your plugin code is loaded. I even consider it a bad practice to add actions or hooks directly as your code is loaded. Deferring everything to `plugins_loaded` is much better. It gives others one place to completely short-circuit your code, and it ensures that they can reliably interact with your hooks.

marcwieland95:
For now, most hooks are not working at all.

This is a major issue in my eyes.

jacobbednarz:
@marcwieland95 @markjaquith are either of you open to providing a PR for what you'd like to see change here? I'm happy to review/test it but my knowledge of how this expected to work is pretty limited in the WordPress ecosystem.

marcwieland95:
I had a quick look into it when I had the issue. But it doesn't seem too simple since there are multiple classes fired at the wrong time (as far as I remember).

I can't guarantee if and when I can make time to create a PR. But since this is a big issue I should be definitely solved ;-) I created a super ugly workaround, so in my projects, it's working for now.

weezer311:
I too am having an issue with this. I'm trying to hook update_option.

ivptr:
Please be so kind to reopen this issue.

@marcwieland95 What does your workaround look like?

marcwieland95:
Here you go. Not proud, but it's solving this issue since I created it 2.5 years ago.

```
protected static $purgeActions = [
        'deleted_post', // Delete a post
        'edit_post', // Edit a post - includes leaving comments
        'delete_attachment', // Delete an attachment - includes re-uploading
        'post_updated', // Update a post
        'comment_post', // Post a comment
        'acf/save_post', // ACF Data saved (including option pages)
];

public function __construct()
{
    // This is a super hacky way to trigger a hook from a third-party plugin which is preregistered
    foreach (static::$purgeActions as $action) {
        add_action($action, [$this, 'hackyForceClearCache'], 80);
    }
}

public function hackyForceClearCache()
{
    do_action('autoptimize_action_cachepurged');

    if (WP_DEBUG) {
        error_log('CloudFlare Cache purged (timestamp: ' . time());
[...]

PR #411, opened 2021-05-21
Load plugin logic within action so filters actually can be triggered

marcwieland95:
Hi @jacobbednarz

As in #345 discussed here's the PR for the hook issue. It actually was way easier than I thought ;-)

Would be nice if you could double-check if everything is still working file and the API calls are actually made.

Thanks and Cheers
Marc

### DIFF
--- a/cloudflare.php
+++ b/cloudflare.php
@@ -46,4 +46,6 @@ if (version_compare(PHP_VERSION, CLOUDFLARE_MIN_PHP_VERSION, '<')) {
 
 // Plugin uses namespaces. To support old PHP version which doesn't support
 // namespaces we load everything in "cloudflare.loader.php"
-require_once CLOUDFLARE_PLUGIN_DIR.'cloudflare.loader.php';
+add_action('init', function () {
+    require_once CLOUDFLARE_PLUGIN_DIR . 'cloudflare.loader.php';
+});


jacobbednarz:
@sejoker could you please also take a peak? I'll get this running locally to confirm as well.

jacobbednarz:
after improving the dev experience in #413, i've been running this locally and hitting an exception for all interactions that make a HTTP call.

> CSRF Token not found. It's possible another plugin is altering requests sent by the Cloudflare plugin.

LCosPEDb-OeTWZTuP-N3gqo07y

the exception is being raised from https://github.com/cloudflare/Cloudflare-WordPress/blob/b12a70697c167a10d7bedb93a16c1c9cb16913c4/src/WordPress/Proxy.php#L67 if you like to dig into it further.

cloudflare.php
```diff
@@ -46,4 +46,6 @@
 
 // Plugin uses namespaces. To support old PHP version which doesn't support
 // namespaces we load everything in "cloudflare.loader.php"
-require_once CLOUDFLARE_PLUGIN_DIR.'cloudflare.loader.php';
+add_action('init', function () {
```
lkraav, line 49: At the end of the day, this should probably be `plugins_loaded`. 

1. In general, we shouldn't be aiming for compatibility for theme-based overrides in plugins - rather optimize for recommending mu-plugins.

2. Theme-based override compatibility can also be achieved by simply not doing "too early for themes" things in the loader, but that probably requires extra refactoring. Workload vs benefits question.

marcwieland95: I think a lot of people try to alter hooks from a theme. It's pretty rare that people use so called site plugin to change functionality and just use a theme just for styling.

In my eyes, this is critical to make it work within themes.

What's the issue in delaying the initiation of the full Cloudflare loader?

lkraav: > What's the issue in delaying the initiation of the full Cloudflare loader?

It's true, in general, `init` should be fine, too. But https://github.com/cloudflare/Cloudflare-WordPress/pull/411#issuecomment-851162263 seems indicate it might be too late, so perhaps additional refactoring is already required with this move.

weezer311: Loading it at `init` works for my use case. Adding a filter for `cloudflare_purge_url_actions` in the theme.

weezer311: Can this be reviewed? It may not solve all of the issues but it will solve some.

jacobbednarz:
closing this off due to becoming stale. 

if someone would like to revive this, please open a fresh PR off the latest master branch and i'll take a look.

PR #446, opened 2021-11-11
Remove Guzzle in favor of WordPress core functions

### LINKED ISSUE #445: 4.7.0 Guzzle dependency breaks sites using different versions
### Confirmation

### WordPress version

5.7.3

4.7.0

### PHP version

7.4.22

### Expected result

Guzzle versions shouldn't collide when using Composer to manage PHP dependencies. If a site is using guzzle, maybe Cloudflare should suggest that its plugin requires it and let the user install the dependency on his own.

### Actual result

Due to the way autoloading is done, the guzzle version from Cloudlare's plugin is the one that is left available for all code to use, leading to issues on how other plugins use it.

### Steps to reproduce

1. Add a custom dependency to Composer that requires Guzzle
2. Use something in the dependency that requires code not present in version ~5.0 (~7.0, for example)
2. Require Cloudflare's plugin
3. Check if the feature dependency in 1. is working

### Additional factoids

It's already mentioned, but it might be important to point out again. We're using Composer to manage our dependencies, both internal/custom and external (packgist and wpackagist).

### References

_No response_

ethanclevenger91:
Closes #445 

This PR is _probably_ incomplete. A couple notes:

- Moves away from Guzzle exception classes. Replaced with what I considered sensible alternatives, but this will alter the logs that result from failed requests
- Inexperienced with stubbing WordPress core functions, so you may desire more comprehensive testing
- Docker image wouldn't build w/o adding an argument to `apt-get update`

But that said, all tests pass and I'll be giving this a go on a staging client site.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 *.DS_Store
 vendor/
 xdebug/*
+.phpunit.result.cache
--- a/Dockerfile.wordpress
+++ b/Dockerfile.wordpress
@@ -1,5 +1,5 @@
 FROM wpdiaries/wordpress-xdebug:5.7-php7.4-apache
-RUN apt-get update && apt-get install -y git-core vim
+RUN apt-get --allow-releaseinfo-change update && apt-get install -y git-core vim
 RUN curl -O https://raw.githubusercontent.com/wp-cli/builds/gh-pages/phar/wp-cli.phar \
     && chmod +x wp-cli.phar \
     && mv wp-cli.phar /usr/local/bin/wp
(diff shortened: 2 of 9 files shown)


ethanclevenger91:
Oof, yeah, looking at the commit history, looks like some of this logic is part of some alternative centralized bunch of code that gets split between all your various plugins, so this probably isn't going to work super well from a workflow perspective, though it seems to be working fine standalone.

jacobbednarz:
Thanks for this @ethanclevenger91! Luckily, I had combined `cloudflare-backend` not too long ago, so there aren't any additional plugins that rely on this now. The frontend is still separate however, there are some plans in the works to clear that up as well once we find some time for it.

Testing this locally looks pretty good to me and I can't stop anything that is out of what we would expect (even with exceptions). I'm happy to land this and if anything else pops up from it, we just address it in an upcoming release.

jacobbednarz (approved):

PR #454, opened 2022-03-13
More efficient cache purge 

### LINKED ISSUE #429: Purge attempted for external URL (feeds.feedburner.com)
### Confirmation

### WordPress version

5.7.2

4.5.1

### PHP version

7.4

### Expected result

No purge errors related to external URLs. Purging should not be attempted for external URLs.

### Actual result

```
[31-Jul-2021 01:40:33 UTC] [Cloudflare] ERROR: [CLIENT API] Array
(
    [type] => response
    [body] => Array
        (
            [result] => 
            [success] => 
            [errors] => Array
                (
                    [0] => Array
                        (
                            [code] => 1082
                            [message] => Unable to purge feeds.feedburner.com. You can only purge files for this zone.
                        )

                    [1] => Array
                        (
                            [code] => 1016
                            [message] => One or more cache purge errors
                        )

                )

            [messages] => Array
                (
                )

        )

)
```

### Steps to reproduce

1. In the Genesis Theme settings, redirect the feed URL to feeds.feedburner.com - this used to be a feature, and the result is now saved in `wp_options` in the option `genesis-settings`.
2. Purge the cache

### Additional factoids

I use WP Fastest Cache with Cloudflare integration, but the `[Cloudflare]` prefix makes me think the error is coming from the Cloudflare plugin. If not, please advise.

### References

_No response_

### LINKED ISSUE #448: 4.7.0: cache purges triggered also on non-public post type changes === massive amount of dead-end work
### Confirmation

### WordPress version

5.8.2

4.7.0

### PHP version

7.4

### Expected result

We are running an e-learning platform with gamification elements, where many post types are non-public, cannot be browsed on frontend.

Cloudflare plugin should not trigger cache purges on any of such post types status transitions.

### Actual result

We fail to check whether post type object is even queryable on frontend, and trigger cache purge operations on URLs that are never accessible on frontend. This results potentially in a metric ton of dead end work.

### Steps to reproduce

1. Use `register_post_type()` with args including either `public = false` or `publicly_queryable => false` (I'm still researching which is more useful here) https://developer.wordpress.org/reference/functions/register_post_type/#parameters
2. Create a new post in this post type.
3. Watch CF plugin trigger cache purge.

### Additional factoids

We need to add additional conditional work at https://github.com/cloudflare/Cloudflare-WordPress/blob/55017a3c33ac3db6183ca6fbfe7a64e892e27c1b/src/WordPress/Hooks.php#L139-L147

Probably something with https://developer.wordpress.org/reference/functions/get_post_type_object/

```php
...
$postType = get_post_type_object( get_post_type( $postId ) );
if (isset($postType->publicly_queryable) && ! $postType->publicly_queryable) {
    return;
}
...
```

### References

N/A

jacobbednarz:
This PR focuses on improving the accuracy and number of URLs that we end up 
sending to the cache purge service. Included are three improvements:

**Don't attempt to purge URLs outside of the current zone**

Cache purge requests are tied to credentials and if those credentials don't 
match the current zone, the URL is thrown away. 

**Don't purge non-public facing posts** 

Many systems have a concept of "public facing" content; this is content that is 
designed for end users. Previously, we didn't take into account things like 
backend URLs, posts that were private and content that wouldn't ever be in the 
caches. With this change, we'll rely on the WordPress context to determine if
the content is publicly queryable and subsequently, purgable.

**Only purge cachable file extensions unless we find hints in page rules or APO 
is enabled**

By default, caching within Cloudflare is restricted to a handful of file
[extensions][1]. However, within the plugin, we just pass it all the
cache purge irrespective of the file extension or zone configuration. This
means we're sending a boat load more URLs to the service than what we will
never be purging.

To combat this we are introducing some additional logic that checks two things:

Firstly, we look at the page rules for the presence of any cache override 
behaviour should it not be present, filter the URLs to only purge those in the 
default caching extensions. This should be considered quite safe as out of the 
box, everyone starts with this configuration and is only overridden when someone 
explicitly sets this in page rules. We don't bother reimplementing the page rule 
globbing logic but instead just check for the presence of "cache_everything" 
for the cache level in any rule.

Secondly, we ensure the zone doesn't have APO enabled. APO is a feature that 
relies on custom purge logic we don't want to duplicate here so just skip it if 
it exists.

To provide a more visual overview, here is a quick diagram of the flow with 
these proposed updates.

```
┌─────────────────────┐
│    purge request    │
│      initiated      │
└─────────────────────┘
           │
           │
           ▼
┌─────────────────────┐
│ post is auto saved  │
│  or post revision?  │───yes───▶ do nothing, exit
└─────────────────────┘
           │
          no
           │
           ▼
┌─────────────────────┐
│   post is public?   │────no───▶ do nothing, exit
└─────────────────────┘
           │
          yes
           │
           │
           ▼
┌ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ┐
    runs cloudflare_purge_by_url hook
└ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ┘
           │
           │
           ▼
┌─────────────────────────────────────────────┐
│  zone has cache_level = "cache_everything"  │
│     page rule present OR is using APO?      │
└─────┬─────────┬─────────────────────────────┘
      │         │
      │         │
      │         │
      │         │    ┌────────────────────────┐
      │         │    │  remove any URLs that  │
      │         │    │         don't          │
      │         └───▶│  have a cachable file  │
      │              │       extension        │
      │              └────────────────────────┘
      │                           │
      │                           │
      │      ┌────────────────────┘
      │      │
      ▼      ▼
 ┌──────────────────────────┐         ┌──────────────┐
 │URL matches current zone? │───no───▶│ discard URL  │
 └──────────────────────────┘         └──────────────┘
               │
              yes
               │
               ▼
 ┌──────────────────────────┐
 │    purge request sent    │
 └──────────────────────────┘
```

Worth noting as well, these restrictions **do not apply** to the UI 
where you can provide URLs (that is separate to this).

hvi1Cy62-NpJ37sJD-i6CGv0Ao

Closes #448
Closes #429

[1]: https://developers.cloudflare.com/cache/about/default-cache-behavior/#default-cached-file-extensions

### DIFF
--- a/src/WordPress/Hooks.php
+++ b/src/WordPress/Hooks.php
@@ -20,6 +20,16 @@ class Hooks
     const CLOUDFLARE_JSON = 'CLOUDFLARE_JSON';
     const WP_AJAX_ACTION = 'cloudflare_proxy';
 
+    // See https://developers.cloudflare.com/cache/about/default-cache-behavior/
+    const CLOUDFLARE_CACHABLE_EXTENSIONS = [
+        "7z", "csv", "gif", "midi", "png", "tif", "zip", "avi", "doc", "gz",
+        "mkv", "ppt", "tiff", "zst", "avif", "docx", "ico", "mp3", "pptx",
+        "ttf", "apk", "dmg", "iso", "mp4", "ps", "webm", "bin", "ejs", "jar",
+        "ogg", "rar", "webp", "bmp", "eot", "jpg", "otf", "svg", "woff", "bz2",
+        "eps", "jpeg", "pdf", "svgz", "woff2", "class", "exe", "js", "pict",
+        "swf", "xls", "css", "flac", "mid", "pls", "tar", "xlsx"
+    ];
+
     public function __construct()
     {
         $this->config = new Integration\DefaultConfig(file_get_contents(CLOUDFLARE_PLUGIN_DIR . 'config.json', true));
@@ -155,6 +165,14 @@ class Hooks
             $urls = apply_filters('cloudflare_purge_by_url', $urls, $postId);
 
             $zoneTag = $this->api->getZoneTag($wpDomain);
+
+            // Fetch the page rules and should we not have any hints of cache
+            // all behaviour or APO, filter out the non-cacheable URLs.
+            $hasCacheOverride = $this->pageRuleContainsCacheEverythingAction($this->api->getPageRules($zoneTag, "active"));
+            if (!$hasCacheOverride && !$this->isAutomaticPlatformOptimizationEnabled()) {
+                $this->logger->debug("cache everything behaviour and APO not found, filtering URLs to only be those that are cacheable by default");
+                $urls = array_filter($urls, array($this, "pathHasCachableFileExtension"));
+            }
 
             // Don't attempt to purge anything outside of the provided zone.
             foreach ($urls as $key => $url) {
--- a/src/WordPress/Hooks.php
+++ b/src/WordPress/Hooks.php
@@ -22,7 +22,7 @@ class Hooks
 
     public function __construct()
     {
-        $this->config = new Integration\DefaultConfig(file_get_contents(CLOUDFLARE_PLUGIN_DIR.'config.json', true));
+        $this->config = new Integration\DefaultConfig(file_get_contents(CLOUDFLARE_PLUGIN_DIR . 'config.json', true));
         $this->logger = new Integration\DefaultLogger($this->config->getValue('debug'));
         $this->dataStore = new DataStore($this->logger);
         $this->integrationAPI = new WordPressAPI($this->dataStore);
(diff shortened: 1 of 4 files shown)


src/WordPress/Hooks.php
```diff
@@ -20,9 +20,19 @@ class Hooks
     const CLOUDFLARE_JSON = 'CLOUDFLARE_JSON';
     const WP_AJAX_ACTION = 'cloudflare_proxy';
 
+    // See https://developers.cloudflare.com/cache/about/default-cache-behavior/
+    const CLOUDFLARE_CACHABLE_EXTENSIONS = [
```
manatarms, line 24: A potential improvement, I wonder if this data is exposed via a URL, in which case we can get an extensions list at build time.

jafowler (approved):

src/WordPress/Hooks.php
patryk: They won't be changing often. It's safe enough to update them along with releases.

PR #466, opened 2022-04-06
Fix for APO not purging cache on page save

Dalias96:
Saving changes on a page was not triggering cache purge for that page. Condition for testing used check for publicly_queryable prop, which by default return false for pages. However WP has build-in function for checking if queried object can be publicly viewed and returns correct values for all types of posts.

### DIFF
--- a/src/WordPress/Hooks.php
+++ b/src/WordPress/Hooks.php
@@ -152,7 +152,7 @@ class Hooks
             }
 
             $postType = get_post_type_object(get_post_type($postId));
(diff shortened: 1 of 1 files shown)


jacobbednarz (approved):

anthonyeden:
Hello,

I don't want to nag, but I was wondering if you would consider releasing this fix to the WordPress.org directory soon? This would be breaking cache clearing for a lot of sites.

Thanks,
Anthony.

PR #469, opened 2022-04-25
Allow clearing cache for array of post ids or single post id

midweste:
I have a use case that would have required many extraneous requests to clear related urls from the cache.

Use case: Woocommerce coupon is updated. Products that are valid for this coupon and their related links need to be cleared from cache.

Problem: Cloudflare module allows me to clear related links based on a single post id.  In my case, there is significant overlap between the related posts.  Causes many extra requests to clear urls.

Solution: Allow for either a postId or postIds array to be process by Hooks purgeCacheByRelevantURLs.  Aggregate all related links and apply array_unique to only clear overlapping urls once.

Results:
image

For 24 products (posts) that need their related links cleared, pull request reduces the amount of API requests from 31 to 7 (442% decrease) and the amount of individuals chunked urls that need to be cleared from 902 to 192.  This is for a forced https site, non forced https sites would double that amount.

### DIFF
--- a/src/WordPress/Hooks.php
+++ b/src/WordPress/Hooks.php
@@ -164,7 +164,7 @@ class Hooks
                     continue;
                 }
 
-                $urls = array_values(array_unique(array_merge($urls, $this->getPostRelatedLinks($postId))));
+                $urls = array_merge($urls, $this->getPostRelatedLinks($postId));
                 $urls = apply_filters('cloudflare_purge_by_url', $urls, $postId);
             }
             $urls = apply_filters('cloudflare_purge_by_urls', $urls, $postIds);
@@ -346,7 +346,7 @@ class Hooks
         }
 
         // Clean array if row empty
-        $listofurls = array_filter($listofurls);
+        $listofurls = array_values(array_unique(array_filter($listofurls)));
 
         return $listofurls;
     }
--- a/src/WordPress/Hooks.php
+++ b/src/WordPress/Hooks.php
@@ -173,6 +173,9 @@ class Hooks
                 return;
             }
 
+            // Filter by unique urls
+            $urls = array_values(array_unique($urls));
+
             $zoneTag = $this->api->getZoneTag($wpDomain);
             $activePageRules = $this->api->getPageRules($zoneTag, "active");
 
--- a/src/WordPress/Hooks.php
+++ b/src/WordPress/Hooks.php
@@ -203,7 +203,7 @@ class Hooks
             }
 
             if (!empty($urls)) {
-                do_action('cloudflare_purge_by_urls', $urls, $postIds);
+                do_action('cloudflare_purged_urls', $urls, $postIds);
                 $chunks = array_chunk($urls, 30);
 
                 foreach ($chunks as $chunk) {
--- a/src/WordPress/Hooks.php
+++ b/src/WordPress/Hooks.php
@@ -145,6 +145,10 @@ class Hooks
                 return;
             }
             $wpDomain = $wpDomainList[0];
+            $zoneTag = $this->api->getZoneTag($wpDomain);
+            if (!isset($zoneTag)) {
+                return;
+            }
 
             $postIds = (array) $postIds;
             $urls = [];
@@ -169,14 +173,20 @@ class Hooks
             }
             $urls = apply_filters('cloudflare_purge_by_urls', $urls, $postIds);
 
-            if (!is_array($urls) || empty($urls)) {
+            // Don't attempt to purge anything outside of the provided zone.
+            foreach ($urls as $key => $url) {
+                if (!Utils::strEndsWith(parse_url($url, PHP_URL_HOST), $wpDomain)) {
+                    unset($urls[$key]);
+                }
+            }
+
+            if (empty($urls)) {
                 return;
             }
 
             // Filter by unique urls
             $urls = array_values(array_filter(array_unique($urls)));
 
-            $zoneTag = $this->api->getZoneTag($wpDomain);
             $activePageRules = $this->api->getPageRules($zoneTag, "active");
 
             // Fetch the page rules and should we not have any hints of cache
@@ -193,14 +203,7 @@ class Hooks
                 $urls = array_filter($urls, array($this, "urlIsHTTPS"));
             }
 
-            // Don't attempt to purge anything outside of the provided zone.
-            foreach ($urls as $key => $url) {
-                if (!Utils::strEndsWith(parse_url($url, PHP_URL_HOST), $wpDomain)) {
-                    unset($urls[$key]);
-                }
-            }
-
-            if (isset($zoneTag) && !empty($urls)) {
+            if (!empty($urls)) {
                 $chunks = array_chunk($urls, 30);
 
                 foreach ($chunks as $chunk) {
--- a/src/WordPress/Hooks.php
+++ b/src/WordPress/Hooks.php
@@ -137,7 +137,7 @@ class Hooks
         }
     }
 
-    public function purgeCacheByRelevantURLs($postId)
+    public function purgeCacheByRelevantURLs($postIds)
     {
         if ($this->isPluginSpecificCacheEnabled() || $this->isAutomaticPlatformOptimizationEnabled()) {
             $wpDomainList = $this->integrationAPI->getDomainList();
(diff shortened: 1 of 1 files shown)


midweste:
An added bonus imo, is that the hooks still run so that will allow users to determine if a url should be added or not.  Added in array_unique to remove any duplicates added by user hooks

src/WordPress/Hooks.php
```diff
@@ -146,28 +146,36 @@ public function purgeCacheByRelevantURLs($postId)
...
-                return;
-            }
+                $savedPost = get_post($postId);
+                if (!is_a($savedPost, 'WP_Post')) {
+                    continue;
+                }
 
-            $urls = $this->getPostRelatedLinks($postId);
-            $urls = apply_filters('cloudflare_purge_by_url', $urls, $postId);
+                $urls = array_values(array_unique(array_merge($urls, $this->getPostRelatedLinks($postId))));
```
jacobbednarz, line 167: the array cleanup seems to fit better in the `getPostRelatedLinks` method before it returns as it isn't related to this method; it should only be using the data that comes back from it.

src/WordPress/Hooks.php
```diff
@@ -146,28 +146,36 @@ public function purgeCacheByRelevantURLs($postId)
...
+                $urls = apply_filters('cloudflare_purge_by_url', $urls, $postId);
+            }
+            $urls = apply_filters('cloudflare_purge_by_urls', $urls, $postIds);
 
-            if (empty($urls)) {
+            if (!is_array($urls) || empty($urls)) {
```
jacobbednarz, line 172: given you've instantiated `$urls = []` at the start of this function, how will the `is_array` check ever fail?

src/WordPress/Hooks.php
```diff
@@ -146,28 +146,36 @@ public function purgeCacheByRelevantURLs($postId)
...
+                if (!is_a($savedPost, 'WP_Post')) {
+                    continue;
+                }
 
-            $urls = $this->getPostRelatedLinks($postId);
-            $urls = apply_filters('cloudflare_purge_by_url', $urls, $postId);
+                $urls = array_values(array_unique(array_merge($urls, $this->getPostRelatedLinks($postId))));
+                $urls = apply_filters('cloudflare_purge_by_url', $urls, $postId);
+            }
+            $urls = apply_filters('cloudflare_purge_by_urls', $urls, $postIds);
```
jacobbednarz, line 170: isn't this doubling the work above? we're already iterating the `postIds` but here we're applying the filter...again?

jacobbednarz (changes requested):

src/WordPress/Hooks.php
midweste: Thanks for the review!  I think the cleanup needs to happen after the user is able to edit and possibly return duplicates, empty urls, or even mistakenly an integer or string.  Never know what user may return. That being said, the loop should probably continue if the user doesn't return an array.

src/WordPress/Hooks.php
midweste: See above, user may return anything from the filter

src/WordPress/Hooks.php
midweste: This is a different hook that has aggregated urls and all post ids, more like a job hook vs the singular cloudflare_purge_by_url that is ran for each post id. Would allow for user to check against all post ids and urls

src/WordPress/Hooks.php
jacobbednarz: i think there may be some confusion here, i'm not saying the functionality shouldn't exist, just that it should live inside `getPostRelatedLinks` before it returns given that is what is acting on the data. 

> Never know what user may return

this doesn't do any sanitization, only merging and de-duping values so i'm not sure what that comment is related to.

src/WordPress/Hooks.php
jacobbednarz: from what i can see, these the same functionality. they both act on `$postIds` - one happens in a loop and the outside the loop. the one inside the loop gets the related links but the other just applies the filter to all of the remain URLs again. i suspect if you remove the second one here, your functionality won't change.

src/WordPress/Hooks.php
midweste: Understood, thank you.  I'll try to address this today if I have some bandwidth.

src/WordPress/Hooks.php
midweste: I did leave the second aggregate hook in the latest commit. Efficiency wise, if you wanted to know the full result of what was about to be cleared and which post_ids were involved, without "cloudflare_purge_by_urls" you would need to monitor the singular cloudflare_purge_by_url for possibly many many iterations (post ids) and save that data to a temporary location (session, global, db, etc) in order to achieve what you could achieve with a single filter and single routine.

src/WordPress/Hooks.php
midweste: an example of what a user may do:

add_filter('cloudflare_purge_by_urls', function($urls) {
   return 'anything_thats_not_an_array';
});

src/WordPress/Hooks.php
midweste: I'm looking this over again.  Moving the unique filtering inside of getPostRelatedLinks does make sense.  However, if you dont apply unique again after you merge with the previous post id related, you end up passing a lot of duplicates to the hook cloudflare_purge_by_url.  I don't think any user would expect that

src/WordPress/Hooks.php
```diff
@@ -326,20 +334,20 @@ public function getPostRelatedLinks($postId)
...
-                array_unique(array_filter($attachmentUrls))
+                $attachmentUrls
             );
         }
 
+        // Clean array and get unique values
+        $listofurls = array_values(array_filter(array_unique($listofurls)));
+
         // Purge https and http URLs
         if (function_exists('force_ssl_admin') && force_ssl_admin()) {
```
midweste, line 345: I was looking over this doubling of urls based on ssl.   Seems to me it would make sense to add this bit of code to the getPostRelatedLinks.  Any reason to add all the http links only to possibly remove them later?

```
 $hasAlwaysUseHTTPSOverrideDisabled = $this->pageRuleContains($activePageRules, "always_use_https", "off");
  if ($this->zoneSettingAlwaysUseHTTPSEnabled($zoneTag) && !$hasAlwaysUseHTTPSOverrideDisabled) {
      $this->logger->debug("always_use_https is enabled without page rule overrides present, removing HTTP based URLs");
      $urls = array_filter($urls, array($this, "urlIsHTTPS"));
  }
```

jacobbednarz: this is to ensure that even after filters are applied, we are cleaning up the HTTP only URLs that will never get cached.

src/WordPress/Hooks.php
jacobbednarz: i'm not sure i follow. can you try explaining another way? 

i still don't think we need both of these `apply_filters` call; we do it inside of the loop and additionally afterwards as well. i cannot see why we would need to duplicate this given every request will pass through the loop and already fetch the related links.

midweste: A common case may be where you clear the local cache when clearing cf cache pages? I e. wp_cache_flush. When it was a single call, the existing would work well, with multiple ids, we would end up clearing the local cache N times. I'm down to remove it, but it does seem that a single hook could be useful.  I'll try to consider some other cases

midweste: Thinking about it more, most of the cases i can think of would be better suited for an action after the cache is actually cleared via api requests

midweste:
@jacobbednarz And by all means, keep commenting and offering direction. I'm far more interested in code that is robust vs quickly published.

src/WordPress/Hooks.php
```diff
@@ -185,14 +202,8 @@ public function purgeCacheByRelevantURLs($postId)
...
-            // Don't attempt to purge anything outside of the provided zone.
-            foreach ($urls as $key => $url) {
-                if (!Utils::strEndsWith(parse_url($url, PHP_URL_HOST), $wpDomain)) {
-                    unset($urls[$key]);
-                }
-            }
-
-            if (isset($zoneTag) && !empty($urls)) {
+            if (!empty($urls)) {
+                do_action('cloudflare_purge_by_urls', $urls, $postIds);
```
jacobbednarz, line 206: what is the thinking here? i don't think we want this here. by this point, we have all the URLs we care about and should just be processing them in chunks.

midweste: ### that one is an action, not a filter. would be very useful here to use to clear local caches for given urls or completely considering CF will soon be requesting then again from origin

jacobbednarz: perhaps a more appropriate name in that case would be something like `cloudflare_purged_urls` to denote it is URLs we are purging?

jacobbednarz:
other than the small naming change, i think we're good here. thanks for the contribution!

src/WordPress/Hooks.php
jacobbednarz, line 206:
```suggestion
                do_action('cloudflare_purged_urls', $urls, $postIds);
```

midweste:
> other than the small naming change, i think we're good here. thanks for the contribution!

Thanks for helping me through it!

PR #476, opened 2022-05-10
Additional checks for autosave and non published posts

midweste:
I will pull the post status checks out of the other two pull requests, with the thought that this one will get addressed first, as the other two depend on it.

### DIFF
--- a/src/WordPress/Hooks.php
+++ b/src/WordPress/Hooks.php
@@ -159,7 +159,7 @@ class Hooks
                 }
 
                 // Do no purge for non-published posts (auto-draft, draft)
-                if (get_post_status($postId) !== 'publish') {
+                if (!in_array(get_post_status($postId), ['publish', 'private', 'password'])) {
                     continue;
                 }
 
--- a/src/WordPress/Hooks.php
+++ b/src/WordPress/Hooks.php
@@ -154,7 +154,12 @@ class Hooks
             $urls = [];
             foreach ($postIds as $postId) {
                 // Do not purge for autosaves or updates to post revisions.
-                if (wp_is_post_autosave($postId) || wp_is_post_revision($postId)) {
+                if (wp_is_post_autosave($postId) || wp_is_post_revision($postId) || (defined('DOING_AUTOSAVE') && DOING_AUTOSAVE)) {
+                    continue;
+                }
+
+                // Do no purge for non-published posts (auto-draft, draft)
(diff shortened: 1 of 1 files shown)


midweste:
You had asked about "private" post status.  According to WP "Private: 

This option hides the content from the public completely.".   I would think then, this should be purged from the cache considering public users cannot view it.

Private content is published only for your eyes, or the eyes of only those with authorization permission levels to see private content. Normal users and visitors will not be aware of private content. It will not appear in the article lists. If a visitor were to guess the URL for your private post, they would still not be able to see your content. You will only see the private content when you are logged into your WordPress blog.

What do you think there? It seems that this type of content should not be cached and should be served dynamically from origin.  Since you can only see these pages while logged in, I believe there is already a mechanism in the plugin that serves content dynamically to logged in users.

I'm going to look at the visibility settings now just as a sanity check

jacobbednarz:
> Private content is published only for your eyes, or the eyes of only those with authorization permission levels to see private content. Normal users and visitors will not be aware of private content. It will not appear in the article lists. If a visitor were to guess the URL for your private post, they would still not be able to see your content. You will only see the private content when you are logged into your WordPress blog.

This is a use case that people are using the plugin for. IIRC, it was a learning platform that uses levels to allow access to the content so yes, private content needs to be considered here.

midweste:
> > Private content is published only for your eyes, or the eyes of only those with authorization permission levels to see private content. Normal users and visitors will not be aware of private content. It will not appear in the article lists. If a visitor were to guess the URL for your private post, they would still not be able to see your content. You will only see the private content when you are logged into your WordPress blog.
> 
> This is a use case that people are using the plugin for. IIRC, it was a learning platform that uses levels to allow access to the content so yes, private content needs to be considered here.

I did add the additional statuses of 'private' and 'password' but after looking more, I'm not sure would make any difference

Per the method initAutomaticPlatformOptimization, the cache is disabled for logged in users.  Private are only ever shown to logged in users, and with this method being run on init, I cant see how private posts could ever been cached.

```
    public function initAutomaticPlatformOptimization()
    {
        // it could be too late to set the headers,
        // return early without triggering a warning in logs
        if (headers_sent()) {
            return;
        }

        // add header unconditionally so we can detect plugin is activated
        if (!is_user_logged_in()) {
            header('cf-edge-cache: cache,platform=wordpress');
        } else {
            header('cf-edge-cache: no-cache');
        }
    }
```

PR #477, opened 2022-05-11
Add filter to allow developers to determine cache usage

midweste:
Use case: Woocommerce customers with role customer should also be served cached pages from cloudflare.  Added a hook to allow the cache state boolean to be manipulated.

### DIFF
--- a/src/WordPress/Hooks.php
+++ b/src/WordPress/Hooks.php
@@ -421,7 +421,8 @@ class Hooks
         }
 
         // add header unconditionally so we can detect plugin is activated
-        if (!is_user_logged_in()) {
+        $cache = apply_filters('cloudflare_use_cache', !is_user_logged_in());
+        if ($cache) {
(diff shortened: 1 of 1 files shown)


midweste:
Our use case:

```
add_filter('cloudflare_use_cache', function ($cache) {
    if (
        is_user_logged_in() &&
        !is_admin() &&
        (function_exists('is_cart') && !\is_cart()) &&
        (function_exists('is_checkout') && !\is_checkout())
    ) {
        $user = wp_get_current_user();
        $roles = (array) $user->roles;
        if (in_array('customer', $roles)) {
            return true;
        }
    }
    return $cache;
});
```

midweste:
How can I get this to run the check again, it seems to be a build test environment error, not a code error

jacobbednarz (approved):

jacobbednarz:
thanks for this one @midweste 🏆

PR #479, opened 2022-05-12
Added cache purge after nav menu edit

Dalias96:
Fixes issue 478

### DIFF
--- a/src/WordPress/Hooks.php
+++ b/src/WordPress/Hooks.php
@@ -158,7 +158,7 @@ class Hooks
                     continue;
                 }
                 
-                // Moved up so $sabedPost can be used earlier.
+                // Moved up so $savedPost can be used earlier.
                 $savedPost = get_post($postId);
                 if (!is_a($savedPost, 'WP_Post')) {
                     continue;


lkraav:
Did you look for any specific hooks "after nav menu saved"?

Micro-managing through menu item posts is almost certainly not the way to go here.

Dalias96:
I was pretty sure that i looked into WP's codex for that kind of hook, but i didn't find right one, and now i looked again and there indeed is "wp_update_nav_menu" hook, which could be the one for this job. I will try to use it.

Dalias96:
@lkraav I added action to `$cloudflarePurgeEverythingActions` its much better solution then original. Thanks for pointing that!

Dalias96:
Actually, what about adding a setting to the settings panel for passing custom hooks, which should trigger CF purge?

jacobbednarz:
i don't think this is what you really want. purging everything is a pretty big deal for some customers that have tens of thousands of posts and cache evictions to happen. i think we need a hook/filter for you to specifically opt into instead to trigger from your side.

rmens:
If a nav/footer menu is edited, that probably has effect on every single page. There might be users who have custom menus that only show up on a subset of pages. Starting with a purge everything and keeping an eye on complaints from edge-cases sounds like the right path to me to get this PR moving.

lkraav:
> If a nav/footer menu is edited, that probably has effect on every single page.

I agree. CF plugin should perhaps post an admin message when full cache purge is triggered - this enables people connect the dots quicker in case something does go wrong performance wise.

rmens:
> CF plugin should perhaps post an admin message when full cache purge is triggered

That's a good suggestion. A lot of other plug-ins like Litespeed cache do this.

Issue #484, opened 2022-06-04
Cloudflare APO Cache Purge not working on the latest version


nsgoyat:
### Confirmation

- [X] My issue isn't already found on the issue tracker.
- [X] I have replicated my issue using the latest version of the plugin and it is still present.

### WordPress version

6.0

### Cloudflare-WordPress version

4.10.0

### PHP version

8.1

### Expected result

The cache should be purged from the Cloudflare servers when we update the site content.

### Actual result

While using the latest version of the plugin and APO, if we update the site content, the Cloudflare Cache isn't purged and it still shows the old content.

### Steps to reproduce

1. Setup the latest Cloudflare Plugin on WordPress.
2. Enable APO settings.
3. Add a Page or article and get it cached on the Cloudflare Servers.
4. Update the article content from the WordPress dashboard and then open the URL in a private window of your browser.
5. You will notice that cache isn't purged from Cloudflare and it still serving the old content.

### Additional factoids

APO Cache Purging works fine when we switch the plugin to version 4.9.1

### References

It happened in the past too: https://github.com/cloudflare/Cloudflare-WordPress/issues/456

zaidoon1:
@nsgoyat can you provide an example of URLs that should be purged but are not being purged

jacobbednarz:
the output of WP_DEBUG would also be helpful here to diagnose whether this is intentional.

jacobbednarz:
just a note from #456, nothing changed functionality wise. the updated release was a fix for the user agent.

nsgoyat:
Hey Folks,
I Have captured the screencast showing the issue. With the latest version, the cache isn't getting purged. When I roll back to the older version, cache purge works fine.

https://photos.app.goo.gl/DiNG6wy9MadsfmJw5

jacobbednarz:
thanks but we need the URLs and debug information here.

ediong:
The concerns raised by this poster is correct. How do I know? I own a news website today.ng and updated the plugin to 4.10.0 about two days ago. I noticed since yesterday that my homepage and other pages were not updating immediately until after a few hours (about six hours or more). I spent most of the time trying to figure out the problem until I stumbled upon this link (https://community.cloudflare.com/t/latest-cloudflare-plugin-update-cache-issues/388638) in the community. It led me to think page and I downgraded as suggested by the poster. Everything is now back to normal. I think the problem is with the latest plugin and it should be looked into asap.

[...]

PR #485, opened 2022-06-05
fix logic for skipping purging unnecessary feed urls

zaidoon1:
current logic filters out NON feed urls:

Screen Shot 2022-06-05 at 3 42 12 PM

while we want to filter out unnecessary feed urls and keep everything else:

Screen Shot 2022-06-05 at 3 41 15 PM

This fixes the issue reported in https://github.com/cloudflare/Cloudflare-WordPress/issues/484


jacobbednarz:
I'll cut this into a hotfix in the coming hour or so when I'm in front of a computer.

jacobbednarz (approved):

PR #489, opened 2022-06-09
Updates purgeCacheByRelevantURLs filter to allow arrays

### LINKED ISSUE #488: Cannot clear cache of a URL defined as an array
### Confirmation

### WordPress version

5.9.3

4.9.1

### PHP version

7.4.29

### Expected result

When adding or updating a post, I expect all URL caches to be cleared, including URLs defined as an array which have been added in, e.g. through the `cloudflare_purge_by_url` hook, or some other method.

### Actual result

URLs which are defined as an array with headers are filtered out, meaning that when a new post is added or updated, services which consume our wordpress content do not have their cache cleared, and stale content continues to be served until a manual cache clear.

### Steps to reproduce

1. Use the `cloudflare_purge_by_url` filter to add a URL that needs it's cache cleared.
```
<?php

/**
 * @param array $urls A list of post related URLs
 * @param integer $post_id the post ID that was modified
 */
function external_service_homepage_cloudflare_url_filter($urls, $post_id) {
    $urls[] = array(
        'url' => 'https://website.com/',
        'headers' => array(
            'x-forwarded-host' => 'www.website.com'
        )
    );
    return $urls;

}

add_filter('cloudflare_purge_by_url', 'external_service_homepage_cloudflare_url_filter', 10, 2);

```

2. Note that the new URL exists in https://github.com/cloudflare/Cloudflare-WordPress/blame/a2a56c4523165969ae3b76b40374d027f52126ea/src/WordPress/Hooks.php#L165
3. Note that the URL is now filtered out https://github.com/cloudflare/Cloudflare-WordPress/blame/a2a56c4523165969ae3b76b40374d027f52126ea/src/WordPress/Hooks.php#L181

This now means that any URL defined as an array which is needing a cache clear on an update won't work.

### Additional factoids

I understand that [this recent update](https://github.com/cloudflare/Cloudflare-WordPress/commit/5e93746b318ab4fd2ed49254aefeae9824aae022) has intentionally removed the ability to clear URL caches outside of the current zone. However, this now means that we can't automatically clear the cache on sites which consume the data out of wordpress as it has to be defined as an array due to the headers.

Being able to have external site cache cleared from a post being added or updated is important as we are using the `wp-json` data to load in content from wordpress onto an external site. Therefore being able to refresh both caches is important. 

### References

https://github.com/cloudflare/Cloudflare-WordPress/commit/5e93746b318ab4fd2ed49254aefeae9824aae022

conorriches:
Closes #488 

This PR should fix the issue where the cache for a URL won't get purged if it's defined as an array, for example if there are headers .

### DIFF
--- a/src/WordPress/Hooks.php
+++ b/src/WordPress/Hooks.php
@@ -174,7 +174,11 @@ class Hooks
 
             // Don't attempt to purge anything outside of the provided zone.
             foreach ($urls as $key => $url) {
-                $url_to_test = is_array($url) ? $url['url'] : $url;
+                $url_to_test = $url;
+                if(is_array($url) && !!$url['url'] ){
+                    $url_to_test = $url['url'];
+                }
+
                 if (!Utils::strEndsWith(parse_url($url_to_test, PHP_URL_HOST), $wpDomain)) {
                     unset($urls[$key]);
                 }
--- a/src/WordPress/Hooks.php
+++ b/src/WordPress/Hooks.php
@@ -175,7 +175,7 @@ class Hooks
             // Don't attempt to purge anything outside of the provided zone.
             foreach ($urls as $key => $url) {
                 $url_to_test = $url;
-                if(is_array($url) && !!$url['url'] ){
+                if (is_array($url) && !!$url['url']) {
                     $url_to_test = $url['url'];
                 }
 
--- a/src/WordPress/Hooks.php
+++ b/src/WordPress/Hooks.php
@@ -174,7 +174,8 @@ class Hooks
 
             // Don't attempt to purge anything outside of the provided zone.
             foreach ($urls as $key => $url) {
-                if (!Utils::strEndsWith(parse_url($url, PHP_URL_HOST), $wpDomain)) {
+                $url_to_test = is_array($url) ? $url['url'] : $url;
+                if (!Utils::strEndsWith(parse_url($url_to_test, PHP_URL_HOST), $wpDomain)) {
                     unset($urls[$key]);
                 }
             }


src/WordPress/Hooks.php
```diff
@@ -174,7 +174,8 @@ public function purgeCacheByRelevantURLs($postIds)
 
             // Don't attempt to purge anything outside of the provided zone.
             foreach ($urls as $key => $url) {
-                if (!Utils::strEndsWith(parse_url($url, PHP_URL_HOST), $wpDomain)) {
+                $url_to_test = is_array($url) ? $url['url'] : $url;
```
jacobbednarz, line 177: you'll want to do a keys check here for url in the array otherwise you'll potentially end up with null.

conorriches: Thanks, I've done that and expanded it out to an `if` statement for extra clarity

DanielBrierton:
Hey @jacobbednarz 👋  Any chance of getting a re-review on this? We're having to manually flush cache each time a blog post is published, which thankfully isn't _too_ frequent, but annoying, nonetheless. If it'll be a little while longer, no worries, we can plan in a little work on our end to work around the issue

jacobbednarz:
apologies for the delay; i've been occupied with other work and thinking about this.

i'm thinking this implementation is _too_ specific for the plugin to cater for. within this we create some assumptions (such as the array key being `"url"`) and only checking a single level of depth. which, to me, seems like we shouldn't be doing this in the first place and whatever is calling the hook needs to instead just pass on only the individual URL? i understand you have a pretty bespoke setup but i'm not sure if that should be supported here for all users. the whole purpose of hooks is that the plugin provides an entrypoint to do a thing to avoid bespoke implementations leaking into the plugin itself.

right now, i'm hesitant to accept this as i can't see a use outside of your specific case which i wouldn't want to add (and maintain). am i missing or not quite understanding anything here?

DanielBrierton:
I guess one piece of info missing up to now is where this array format is coming from. It matches the Cloudflare API for purging files by URL, so it would be invalid without a `url` field, and doesn't support nesting - https://api.cloudflare.com/#zone-purge-files-by-url

Perhaps we made an incorrect assumption that this hook would allow us to use this object format and to be fair, the documentation doesn't say this is supported. It worked for us though until this filter was implemented, since at the end it just gets converted to JSON and POSTed to the endpoint. It affects anyone who wants to purge another URL that has a custom cache key as outlined in the above docs.

jacobbednarz:
gotcha, i see what you mean now. this seems to have only been working by accident as this purge process never had the intention of supporting values outside of the array (despite it being supported in the API). however, in the case where we have URLs and HTTP header for a custom cache key, that seems like we should be supporting it and this PR is no longer about bespoke functionality and actually about allowing full use of the API available.

you've convinced me this should be added, thanks! i'll run this through a few test cases and look to merge.

jacobbednarz (approved):

DanielBrierton:
Thanks @jacobbednarz! Much appreciated 🙂

DanielBrierton:
Hey again @jacobbednarz, sorry to bother you. Any idea when this might get released so we can get it into our WP without having to pull it from master?

PR #502, opened 2023-02-07
Check for doing autosave and non published post types

midweste:
Shouldn't clear if doing autosave or if posts are not published.  Hope all is well!

### DIFF
--- a/src/WordPress/Hooks.php
+++ b/src/WordPress/Hooks.php
@@ -154,7 +154,12 @@ class Hooks
             $urls = [];
             foreach ($postIds as $postId) {
                 // Do not purge for autosaves or updates to post revisions.
-                if (wp_is_post_autosave($postId) || wp_is_post_revision($postId)) {
+                if (wp_is_post_autosave($postId) || wp_is_post_revision($postId) || (defined('DOING_AUTOSAVE') && DOING_AUTOSAVE)) {
+                    continue;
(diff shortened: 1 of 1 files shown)


src/WordPress/Hooks.php
```diff
@@ -154,7 +154,12 @@ public function purgeCacheByRelevantURLs($postIds)
             $urls = [];
             foreach ($postIds as $postId) {
                 // Do not purge for autosaves or updates to post revisions.
-                if (wp_is_post_autosave($postId) || wp_is_post_revision($postId)) {
+                if (wp_is_post_autosave($postId) || wp_is_post_revision($postId) || (defined('DOING_AUTOSAVE') && DOING_AUTOSAVE)) {
```
lkraav, line 157: Why is current `wp_is_post_autosave()` alone not enough?

src/WordPress/Hooks.php
```diff
@@ -154,7 +154,12 @@ public function purgeCacheByRelevantURLs($postIds)
...
+                    continue;
+                }
+
+                // Do no purge for non-published posts (auto-draft, draft)
+                if (!in_array(get_post_status($postId), ['publish', 'private', 'password'])) {
```
lkraav, line 162: Not convinced these strategies should be mixed into a single PR.

PR #503, opened 2023-02-07
Purge cache on taxonomy term change

midweste:
Posts urls should be cleared when associated taxonomy term changes

### DIFF
--- a/src/WordPress/Hooks.php
+++ b/src/WordPress/Hooks.php
@@ -6,6 +6,7 @@ use CF\API\APIInterface;
 use CF\Integration;
 use Psr\Log\LoggerInterface;
 use WP_Taxonomy;
+use WP_Query;
 
 class Hooks
 {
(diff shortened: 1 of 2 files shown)


src/WordPress/Hooks.php
```diff
@@ -478,6 +479,31 @@ public function purgeCacheOnNewComment($comment_id, $comment_status, $comment_da
         $this->purgeCacheByRelevantURLs($comment_data['comment_post_ID']);
     }
 
+    public function purgeCacheOnTermChange($ids, $taxonomy, $clean_taxonomy)
+    {
+        $args = array(
+            'posts_per_page' => -1,
```
elliott-stocks, line 485: This is going to introduce performance issues on large sites as it's going to query for _all_ posts.

It feels like this action itself should be optional and limited to the first ~100 or so posts?

midweste: within a certain taxonomy yes.  In is not the most efficient way to query but I can't think of another way to determine associated posts.

I get your thought about top 100 but what is worse, slower execution for a site author, or incorrect pages for sure visitors.

In another pull request, I've somewhat addressed the site author performance by offloading the related URL purge to cron

PR #532, opened 2024-02-09
CUSTESC-36595: Retrieve the Wordpress domain only from the active zone

aseure:
We recently noticed that a user got its zone incorrectly reported as inactive by the Wordpress plugin. After some investigations, it seems that two zones actually existed: an old purged one, and a new active zone. To guarantee that the plugin always retrieve the active one, if any, this commit changes the zone retrieval call to the Cloudflare API by adding a `status=active` query parameter to the request, [as documented here](https://developers.cloudflare.com/api/operations/zones-get).

### DIFF
--- a/src/WordPress/ClientActions.php
+++ b/src/WordPress/ClientActions.php
@@ -44,7 +44,15 @@ class ClientActions
         // We tried to fetch a zone but it's possible we're using an API token,
         // So try again with a zone name filtered API call
         if (!$this->api->responseOk($response)) {
-            $zoneRequest = new Request('GET', 'zones/', array('name' => idn_to_ascii($this->wordpressAPI->getOriginalDomain(), IDNA_DEFAULT, INTL_IDNA_VARIANT_UTS46), array()));
+            $zoneRequest = new Request(
+                'GET',
+                'zones/',
+                array(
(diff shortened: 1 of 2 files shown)


src/WordPress/ClientActions.php
```diff
@@ -44,7 +44,15 @@ public function returnWordPressDomain()
         // We tried to fetch a zone but it's possible we're using an API token,
         // So try again with a zone name filtered API call
         if (!$this->api->responseOk($response)) {
-            $zoneRequest = new Request('GET', 'zones/', array('name' => idn_to_ascii($this->wordpressAPI->getOriginalDomain(), IDNA_DEFAULT, INTL_IDNA_VARIANT_UTS46), array()));
```
jacobbednarz, line 47: we'll also want to update the call above on L42 in case the user is using API keys. iirc, you'll need to modify `$this->request` to include the params (or find the caller and update it there which is probably the more sustainable way forward).

aseure: Following your suggestion, I've only changed the `WordPressClientAPI.php` file, which is the only other place I found performing a `GET zones/` which may benefit from this.

There is a last `GET zones/` call in https://github.com/aseure/Cloudflare-WordPress/blob/5c147468490835c94b5c8dd69952f80d23f30672/src/API/AbstractPluginActions.php#L99 but since it is only a call to check API key and email validity, I did not bother changing it to limit any unknown negative impact.

jacobbednarz (approved):

PR #539, opened 2024-03-06
Support for PHP 8

Korben00:
### Description

This pull request includes updates to transition the project's PHP support to version 8.0. These changes are necessary to ensure the project stays compatible with the latest PHP version, providing improved performance, security, and newer language features. The update involves modifications to composer.json, composer.lock, and readme.txt files, alongside code adjustments in src/Integration/DefaultLogger.php to align with PHP 8 standards.

### Changes

**1.     composer.json and composer.lock:**

- Updated PHP requirement to "php": "8.0".
- Upgraded dependencies to their PHP 8 compatible versions. This includes major version upgrades for packages such as symfony/yaml from ^2.6 to ^5.3, phpunit/phpunit from * to ^9.3, and psr/log from ^1.0 to ^3.0, among others.
- Adjusted platform config setting to PHP 8.0, ensuring consistency across development and production environments.

**2.     src/Integration/DefaultLogger.php:**

- Implemented type declarations and return type void for methods log and debug to comply with the PHP 8 standards.
- Replaced array notation with [] for default parameter values and function return types.
- Introduced Stringable|string for $message parameter type hinting, leveraging PHP 8's union types and promoting more flexible logging capabilities.

**3.     readme.txt:**

- Updated the "Requires PHP" version to 8.0 to inform users and contributors about the new PHP version requirement.

### DIFF
--- a/.github/workflows/php.yml
+++ b/.github/workflows/php.yml
@@ -14,7 +14,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php-versions: ['8.0', '8.1']
+        php-versions: ['8.0', '8.1', '8.2', '8.3']
         experimental: [false]
 
 
--- a/.github/workflows/php.yml
+++ b/.github/workflows/php.yml
@@ -14,11 +14,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php-versions: ['7.2', '7.3', '7.4']
+        php-versions: ['8.0', '8.1']
         experimental: [false]
-        include:
-          - php-versions: '8.0'
-            experimental: true
 
 
     steps:
--- a/composer.json
+++ b/composer.json
@@ -10,18 +10,18 @@
   "type": "wordpress-plugin",
   "require": {
     "cloudflare/cf-ip-rewrite": "^1.0.0",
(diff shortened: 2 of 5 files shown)


.github/workflows/php.yml
```diff
@@ -14,11 +14,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php-versions: ['7.2', '7.3', '7.4']
+        php-versions: ['8.0', '8.1']
```
LeoColomb, line 17:
```suggestion
        php-versions: ['8.0', '8.1', '8.2', '8.3']
```

LeoColomb:
Mentioning @aseure, just in case 😊

Korben00:
Any time to review ?

lkraav:
Indeed, standard practice still today is to keep at least PHP 7.4 in play.

LeoColomb:
@aseure @lkraav Thanks for your replies.
While I'm personally not a big fan of standard practices actually specific to WordPress, I understand the reasoning. 😊

Thus, we've just submitted another PR (#541) which does not touch the platform version but allows working around the compatibility/conflict issue with the dependencies.

Issue #528, opened 2023-12-12
PHP 8.2 Compatibility


fliespl:
### Confirmation

- [X] My issue isn't already found on the issue tracker.
- [X] I have replicated my issue using the latest version of the plugin and it is still present.

### WordPress version

6.4.2

### Cloudflare-WordPress version

4.12.2

### PHP version

8.2.8

### Expected result

Successful plugin activation

### Actual result

$ wp plugin activate cloudflare
PHP Fatal error:  Declaration of CF\Integration\DefaultLogger::debug($message, array $context = []) must be compatible with Psr\Log\AbstractLogger::debug(Stringable|string $message, array $context = []): void in /nas/content/live/forwardsite/wp-content/plugins/cloudflare/src/Integration/DefaultLogger.php on line 42
Fatal error: Declaration of CF\Integration\DefaultLogger::debug($message, array $context = []) must be compatible with Psr\Log\AbstractLogger::debug(Stringable|string $message, array $context = []): void in /nas/content/live/forwardsite/wp-content/plugins/cloudflare/src/Integration/DefaultLogger.php on line 42

### Steps to reproduce

Trying to use plugin with php 8.2.8.

fliespl:
It's a collision with rollbar plugin.

xanghyr:
Hello !
Any news about when this will be fixed ? I'm actualy running into the same issue and can't enable the plugin.

tomkyser:
> Hello ! Any news about when this will be fixed ? I'm actualy running into the same issue and can't enable the plugin.
> 
> edit : I don't have the "rollbar plugin" on my side.

So we fixed this by forcing certain composer dependencies namely psr/log to use different versions that are compatible with the cloudflare plugin's requirements.

I**F you are using composer for your project try:**
`"psr/log": "^1.0.0",`

If that doesn't fix it then you'll need to keep digging: `composer why {insert a desperate guess dependency name} -t`

There is already a PR that would fix this, not sure why they haven't done it yet. https://github.com/cloudflare/Cloudflare-WordPress/pull/541

PR #541, opened 2024-03-25
Allow `psr/log` v3 when using composer

### LINKED ISSUE #445: 4.7.0 Guzzle dependency breaks sites using different versions

### LINKED ISSUE #470: Update psr/log to 3.0
### Confirmation

### WordPress version

Latest

Latest

### PHP version

8.0

### Expected result

I currently have a project that uses `psr/log` but upon updating to 3.0 I'm getting an error thrown from Cloudflare's implementation due to the `::debug` declaration when extending `AbstractLogger`.

I'd like to request that `psr/log` be updated to 3.0 on the plugin to resolve the conflict if possible.

I'm open to doing a PR for this.

### Actual result

```sh
2022/05/03 07:40:24 [error] 11363#11363: *29714057 FastCGI sent in stderr: "PHP message: PHP Fatal error:  Declaration of CF\Integration\DefaultLogger::debug($message, array $context = []) must be compatible with Psr\Log\AbstractLogger::debug(Stringable|string $message, array $context = []): void in /home/example/webapps/example-staging/releases/20220503073807/web/app/plugins/cloudflare/src/Integration/DefaultLogger.php on line 42"
```

### Steps to reproduce

1. Require psr/log into your site/project (e.g. through composer).
2. Get thrown the error above due to declaration incompatibility.

### Additional factoids

_No response_

### References

_No response_

### LINKED ISSUE #490: Dependency on Psr/log breaks sites using different versions
### Confirmation

### WordPress version

6.0

4.10.1

### PHP version

8

### Expected result

A fatal error should not occur.

### Actual result

A fatal error with the following message occurs.

```
Fatal error: Declaration of Psr\Log\NullLogger::log($level, Stringable|string $message, array $context = []): void must be compatible with Psr\Log\LoggerInterface::log($level, $message, array $context = []) in /app/public/wp-content/themes/site.tld/vendor/psr/log/src/NullLogger.php on line 26
```

### Steps to reproduce

1. Install 4.10.1 version of Cloudflare for WordPress.
2. Create a theme which depends on and uses Psr/log 3.0.0

### Additional factoids

_No response_

### References

A similar issue has occured and has been dealt with in https://github.com/cloudflare/Cloudflare-WordPress/issues/445.

LeoColomb:
Alternative solution for #539.
This helps to work around the issue regarding the conflicted dependencies (specifically `Psr\Log` package).
This would allow installing the package on a PHP 8 server via Composer, while keeping the wp.org plugin a PHP 7.x target.

> [!NOTE]
> This change requires the plugin to be published as a standard package on Packagist and then be released with a new tag.
> https://packagist.org/packages/submit

> [!Warning]
> A conditional class definition is unfortunately required to match `psr/log` methods signatures.
> This is not very pretty, but, again, required.
> [`b6e81b9`](https://github.com/cloudflare/Cloudflare-WordPress/pull/541/commits/b6e81b9ba1bd64d3a12ea31c363d119cb3d44bb4)

---
* Closes #539
* Resolves #490
* Resolves #470
* Resolves #445 
* Ref #528
* Closes Cloudflare case 3130507
* Ref https://wordpress.org/support/topic/compatibility-with-php-8-2-20/
* Ref https://wordpress.org/support/topic/php-function-deprecated-php-in-8-2/
* Ref https://wordpress.org/support/topic/w3totalcache-cloudflare-extension-vs-cloudflare-apo-wordpress-plugin/
* Ref https://wordpress.org/support/topic/fatal-error-on-plugin-activation-42/
* Ref https://wordpress.org/support/topic/php-fatal-error-psrlog-php8/

---

cc @Korben00

### DIFF
--- a/cloudflare.loader.php
+++ b/cloudflare.loader.php
@@ -1,6 +1,8 @@
 <?php
 
-require_once __DIR__.'/vendor/autoload.php';
+if (!class_exists('CloudFlare\IpRewrite')) {
+	require_once __DIR__.'/vendor/autoload.php';
+}
 
 use CloudFlare\IpRewrite;
 
--- a/composer.json
+++ b/composer.json
@@ -11,7 +11,7 @@
   "require": {
     "cloudflare/cf-ip-rewrite": "^1.0.0",
     "symfony/polyfill-intl-idn": "*",
-    "psr/log": "^1.0"
+    "psr/log": "^1.0 || ^3.0"
   },
   "require-dev": {
     "symfony/yaml": "~2.6",
@@ -40,5 +40,8 @@
     "allow-plugins": {
       "dealerdirect/phpcodesniffer-composer-installer": true
     }
+  },
+  "extra": {
+    "installer-name": "cloudflare"
   }
 }
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "e8665bcdae6db5846ed6b42aef3e563e",
+    "content-hash": "38f0352fff5fc31da5152c8912eb191e",
     "packages": [
         {
             "name": "cloudflare/cf-ip-rewrite",
--- a/src/Integration/DefaultLogger.php
+++ b/src/Integration/DefaultLogger.php
@@ -6,6 +6,9 @@ use Psr\Log\AbstractLogger;
 use Psr\Log\LogLevel;
 use Psr\Log\LoggerInterface;
 
+# if (Psr\Log version < 2)
+if (class_exists('Psr\Log\Test\DummyTest')):
+
 class DefaultLogger extends AbstractLogger implements LoggerInterface
 {
     private $debug;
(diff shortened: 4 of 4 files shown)


JulienMelissas:
Would be great to get this fixed! We had a few hundred sites go down today updating Cloudflare via Composer thanks to this fun error 😅 (still trying to figure out why the error didn't show itself in staging)

LeoColomb:
@JulienMelissas If you are a business customer of Cloudflare, don't hesitate to open an internal ticket in their support area. If so, feel free to reference our ticket 3130507.

LeoColomb:
@aseure @jacobbednarz @william-woodhead Any possible feedback?

tomkyser:
@aseure @jacobbednarz @william-woodhead any update? I've got a pile of duct tape getting our composer based project to work with this plugin specifically being the issue. What's holding it up?

dev-triboo:
@aseure  @jacobbednarz  @william-woodhead any update?

thedavidthomas:
Can this be merged instead of automatically closing it please?

asha23:
This still needs fixing - Without it, this plugin is useless for some of us running composer dependencies on php 8

PR #546, opened 2024-07-29
Fix empty settings page for users with `manage_options` capability without administrator role

### LINKED ISSUE #544: Empty settings page for users with custom role having the `manage_options` capability
### Confirmation

### WordPress version

6.4.3

4.12.6

### PHP version

7.4.33

### Expected result

A user with a custom role which has the `manage_options` capability, being able to purge the cache through the **Settings → Cloudflare** page.

### Actual result

When clicking the "Cloudflare" admin menu item, an empty page is displayed and `CONFIG_FETCH_ERROR` and `ZONES_FETCH_ERRORS` errors occur in the console.

### Steps to reproduce

1. Create a user with custom role, having the `manage_options` capability (e.g. using the Members plugin; https://wordpress.org/plugins/members/)
2. Visit **Settings → Cloudflare**

### Additional factoids

It appears that the changes from https://github.com/cloudflare/Cloudflare-WordPress/pull/529 are causing the issues (released in version `4.12.3`). The "Cloudflare" admin menu item requires the `manage_options` capability and the WordPress AJAX action `cloudflare_proxy` — which seems needed to load the settings page — is checking for the `administrator` role.

https://github.com/cloudflare/Cloudflare-WordPress/blob/dd13e1509194ee0a15c4f737082d39cdc226ad71/src/WordPress/Hooks.php#L82-L87

https://github.com/cloudflare/Cloudflare-WordPress/blob/dd13e1509194ee0a15c4f737082d39cdc226ad71/src/WordPress/Proxy.php#L56-L60

It might be better to check against the `manage_options` capability in the proxy too, so both will be checking the same requirement to access the settings page.

<hr>

Also, as mentioned in the WordPress developer documentation at https://developer.wordpress.org/reference/functions/current_user_can/, checking against a role instead of a capability using `current_user_can()` is discouraged:

> While checking against particular roles in place of a capability is supported in part, this practice is discouraged as it may produce unreliable results.

https://github.com/cloudflare/Cloudflare-WordPress/blob/dd13e1509194ee0a15c4f737082d39cdc226ad71/src/WordPress/WordPressAPI.php#L159-L165

https://github.com/cloudflare/Cloudflare-WordPress/blob/58db13b91fbd5e8613a8599d58cf05d04914d7e6/src/WordPress/WordPressWrapper.php#L39-L42

### References

https://github.com/cloudflare/Cloudflare-WordPress/pull/529

rvdsteege:
The WordPress admin settings page of the Cloudflare plugin is empty for users having the `manage_options` capability without the administrator role (as is the case with for example a custom 'Manager' role).

This PR intends to fix #544 which describes this issue, by using the same `manage_options` capability check for the admin menu page and the 'Cloudflare proxy':

> The "Cloudflare" admin menu item requires the `manage_options` capability and the WordPress AJAX action `cloudflare_proxy` — which seems needed to load the settings page — is checking for the `administrator` role.

Can you take a look at this, @aseure?

### DIFF
--- a/src/WordPress/WordPressAPI.php
+++ b/src/WordPress/WordPressAPI.php
@@ -161,6 +161,6 @@ class WordPressAPI implements IntegrationAPIInterface
      */
     public function isCurrentUserAdministrator()
     {
-        return $this->wordPressWrapper->currentUserCan('administrator');
+        return $this->wordPressWrapper->currentUserCan('manage_options');
     }
 }


rvdsteege:
@aseure it's been a month, any update on this? 🙂

rvdsteege:
Hey @jacobbednarz, is this something you could merge? I noticed previously mentioned user _aseure_ doesn't work for Cloudflare anymore.

I've been trying for some time to get https://github.com/cloudflare/Cloudflare-WordPress/issues/544 resolved, but I don't get any response here on GitHub nor in Cloudflare support case 01317769.

Issue #554, opened 2025-08-11
Psr\Log prevents other plugins from being activated


davidbarratt:
### Confirmation

- [x] My issue isn't already found on the issue tracker.
- [x] I have replicated my issue using the latest version of the plugin and it is still present.

### WordPress version

6.8.2

### Cloudflare-WordPress version

4.12.8

### PHP version

8.4.10

### Expected result

I expected to be able to activate [Flex HSA/FSA Payments](https://wordpress.org/plugins/pay-with-flex/) alongside Cloudflare without any issues.

### Actual result

A fatal error see https://github.com/JoinFlexHealth/woocommerce/issues/1

### Steps to reproduce

1. Enable Cloudflare
2. Attempt to enable [Flex HSA/FSA Payments](https://wordpress.org/plugins/pay-with-flex/)

### Additional factoids

This appears to be because [psr/log](https://packagist.org/packages/psr/log) is in an incompatible version. I think this could be solved by using the [Jetpack Autoloader](https://developer.jetpack.com/docs/jetpack-development/autoloader/)

### References

- #528
- #490
- #470
- #541
- #539

davidbarratt:
Interesting, so using Jetpack Autoloader results in this error:

> PHP Fatal error:  Declaration of CF\Integration\DefaultLogger::debug($message, array $context = []) must be compatible with Psr\Log\AbstractLogger::debug(Stringable|string $message, array $context = []): void in cloudflare/src/Integration/DefaultLogger.php on line 42

If this plugin wants to use an _older_ version of `Psr\Log` (on an unsupported version of PHP) I feel like the best solution here is to prefix the dependency?

davidbarratt:
Here are the options that I can think of and I would appreciate any feedback from @jacobbednarz (or anyone at @cloudflare)
1. Use the latest version of [psr/log](https://packagist.org/packages/psr/log) and bump support to PHP [8.0.0+](https://www.php.net/releases/8.0/en.php), a version of PHP that entered end-of-life (EOL) [nearly 2 years ago](https://endoflife.date/php).
   1. Also use [Jetpack Autoloader](https://developer.jetpack.com/docs/jetpack-development/autoloader/) to avoid this problem in the future.
2. Use [PHP Scoper](https://github.com/humbug/php-scoper) to [modify the namespace](https://deliciousbrains.com/php-scoper-namespace-composer-depencies/) of the dependencies.
[...]

PR #555, opened 2025-08-20
Compat: add psr/log v3 support in DefaultLogger; align build/deps

lgobatto:
This PR adds compatibility with psr/log v3 in DefaultLogger and adjusts related tests and Composer constraints.

Why:
- Many modern projects and frameworks have already moved to psr/log v3; keeping only v1/v2 compatibility makes integration harder and can cause dependency conflicts.

What changed:
- DefaultLogger updated to be compatible with psr/log v3 while preserving existing behavior.
- Integration tests updated to cover the new interface expectations.
- Composer dependencies adjusted to ensure compatibility with newer environments.

Notes:
- The release workflow file was aligned with upstream to avoid unrelated changes in this PR.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,47 +1,47 @@
 name: Release plugin
 on:
-    release:
-        types: [published]
+  release:
+    types: [published]
 jobs:
-    release:
-        name: New release
-        runs-on: ubuntu-latest
-        steps:
-            - name: Checkout code
-              uses: actions/checkout@v2
+  release:
+    name: New release
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


barrykooij:
At this moment, we can't upgrade `psr/log` to `^3.0`. This would require us to bump our required PHP version to `8.0+` and, unfortunately, the WordPress ecosystem does not seem ready for this yet.

In release `4.14.0` we've added vendor prefixing for all used dependencies. This will resolve this issue and prevent future ones. You can read more about this in this issue: https://github.com/cloudflare/Cloudflare-WordPress/issues/554

Thank you for your pull request, and I'm sorry we couldn't implement it in this way.

